PR #15, opened 2019-09-23
T-REX 1.1

Myna65:


contracts/registry/IdentityRegistry.sol
```diff
@@ -1,40 +1,73 @@
...
-    //Array storing trusted claim types of the security token.
-    uint256[] claimTypes;
 
-    IdentityContract[] public identities;
+    /**
+     * @dev Returns true if the caller is the current owner.
+     */
+    function isAgent() public view returns (bool) {
+        return agents[msg.sender];
+        // return msg.sender == _agent;
```
Myna65:
```suggestion
```

contracts/token/IToken.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.5.10;
+
+//interface
+contract IToken{
+
+    string public name;
+    string public symbol;
+    uint8 public constant decimals = 0;
```
Joachim-Lebrun: i don't think we should set a value for this on the interface, it should be set on the implementation side.

Joachim-Lebrun: https://github.com/TokenySolutions/T-REX/pull/17

Joachim-Lebrun:
@dilipror @Nakasar @Aboudjem @Myna65 @lfalempin @adetante @fcroiseaux  we expect to merge soon, if you can review the code it would be great :)

PR #20, opened 2019-11-15
Update TrustedIssuersRegistry.sol

Aboudjem:
temporary fix by setting false on removal
Trusted Issuer is correctly removed from the trustedIssuers mapping but returning true on isTrustedIssuer()


Nakasar:
Should be fixed by https://github.com/TokenySolutions/T-REX/pull/15/commits/9daf5afaac150fc6cb4924464c25050cd158804a

PR #30, opened 2020-01-16
Pausable transfer functions

dilipror:


Nakasar:
You should also increase the version in the `package.json` to `2.0.4`.

PR #31, opened 2020-02-06
Create a ForcedTransfer function

Joachim-Lebrun:
This function will be used to transfer tokens on behalf of an investor without the need of any transaction on the blockchain coming from the sender. This way the sender doesn't need any ETH on his wallet and doesn't need to manage crypto.

### DIFF
--- a/contracts/token/TransferManager.sol
+++ b/contracts/token/TransferManager.sol
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
 
         revert("Transfer not possible");
     }
-
+    
+    /**
+    * 
+    *  Require that the from and to addresses are not frozen.
+    *  Require that the value should not exceed available balance .
+    *  Require that the to address is a verified address,
+    *  If the `to` address is not currently a shareholder then it MUST become one.
+    *  If the transfer will reduce `from`'s balance to 0 then that address
+    *  MUST be removed from the list of shareholders.
+    *
+    * @param _from The address of the sender
+    * @param _to The address of the receiver
+    * @param _value The number of tokens to transfer
+    *
+    * @return `true` if successful and revert if unsuccessful
+    */
+    function forcedTransfer(address _from, address _to, uint256 _value) onlyAgent external returns (bool) {
+        require(!frozen[_to] && !frozen[_from]);
+        require(_value <=  balanceOf(_from).sub(freezedTokens[_from]), "Sender Has Insufficient Balance" );
+        if(identityRegistry.isVerified(_to) && compliance.canTransfer(_from, _to, _value)){
+            updateShareholders(_to);
+            pruneShareholders(_from, _value);
+            _transfer(_from, _to, _value);
+            return true;
+        }
+        revert("Transfer not possible");
+    }
+    
     /**
      * Holder count simply returns the total number of token holder addresses.
      */
--- a/contracts/token/TransferManager.sol
+++ b/contracts/token/TransferManager.sol
@@ -78,7 +78,7 @@ contract TransferManager is Pausable {
     mapping(address => address) private cancellations;
     mapping (address => bool) frozen;
     mapping (address => Identity)  _identity;
-    mapping (address => uint256) public freezedTokens;
+    mapping (address => uint256) public frozenTokens;
 
     mapping(uint16 => uint256) countryShareHolders;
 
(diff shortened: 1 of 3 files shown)


contracts/token/TransferManager.sol
```diff
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
 
         revert("Transfer not possible");
     }
-
+    
+    /**
+    * 
+    *  Require that the from and to addresses are not frozen.
+    *  Require that the value should not exceed available balance .
```
Nakasar: There is a trailing space here:

```suggestion
    *  Require that the value should not exceed available balance.
```

contracts/token/TransferManager.sol
```diff
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
...
+    *  Require that the to address is a verified address,
```
Nakasar: Coherence with other param description:

```suggestion
    *  Require that the `to` address is a verified address,
```

contracts/token/TransferManager.sol
```diff
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
...
+    *
+    * @param _from The address of the sender
+    * @param _to The address of the receiver
+    * @param _value The number of tokens to transfer
+    *
+    * @return `true` if successful and revert if unsuccessful
+    */
+    function forcedTransfer(address _from, address _to, uint256 _value) onlyAgent external returns (bool) {
+        require(!frozen[_to] && !frozen[_from]);
+        require(_value <=  balanceOf(_from).sub(freezedTokens[_from]), "Sender Has Insufficient Balance" );
```
Nakasar: Trailing space:

```suggestion
        require(_value <=  balanceOf(_from).sub(freezedTokens[_from]), "Sender Has Insufficient Balance");
```

contracts/token/TransferManager.sol
```diff
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
 
         revert("Transfer not possible");
     }
-
+    
+    /**
+    * 
+    *  Require that the from and to addresses are not frozen.
```
Nakasar: Isn't the agent allowed to execute the forced transfer even if addresses are frozen?

contracts/token/TransferManager.sol
```diff
@@ -183,7 +183,34 @@ contract TransferManager is Pausable {
...
+    *  Require that the to address is a verified address,
+    *  If the `to` address is not currently a shareholder then it MUST become one.
+    *  If the transfer will reduce `from`'s balance to 0 then that address
+    *  MUST be removed from the list of shareholders.
+    *
+    * @param _from The address of the sender
+    * @param _to The address of the receiver
+    * @param _value The number of tokens to transfer
+    *
+    * @return `true` if successful and revert if unsuccessful
```
Nakasar: I think you can remove the return boolean. If the function reverts it has failed, otherwise it was a success.

Joachim-Lebrun: well, that's true but the standard `transfer` and `transferFrom` functions in ERC20 are working with a bool return, so i liked the idea to work the same way with `forcedTransfer` even if it doesn't change much indeed

Nakasar: Oh, you're right! Yep, we should follow the same standard!

Nakasar:
I have update the version of the package with a minor release.

Joachim-Lebrun:
> You need to update the wording in other contracts as well, especially `TransferManager` which has the declaration of the mapping:
> 
> ```
> mapping (address => uint256) public freezedTokens;
> ```

done

Joachim-Lebrun:
I still need to write tests for forcedTransfer function before merging the PR

dilipror:
> Dilip still needs to write tests for forcedTransfer function before merging the PR

Done

test/tokenTransfer.test.js
```diff
@@ -829,16 +829,43 @@ contract("Token", accounts => {
...
+  it("Successful forced transfer", async () => {
+    let tx = await token.forcedTransfer(user1, user2, 300, { from: agent })
+      .should.be.fulfilled;
+    log(`Cumulative gas cost for token transfer ${tx.receipt.gasUsed}`);
+
+    let balance1 = await token.balanceOf(user1);
+    let balance2 = await token.balanceOf(user2);
+    log(`user1 balance: ${balance1}`);
+    log(`user2 balance: ${balance2}`);
+  });
```
Nakasar:
```suggestion

    balance1.toString().should.equal('700');
    balance2.toString().should.equal('300');
  });
```

PR #32, opened 2020-02-11
added the possibility to issue transactions in batch

Joachim-Lebrun:


contracts/token/TransferManager.sol
```diff
@@ -157,6 +157,12 @@ contract TransferManager is Pausable {
         revert("Transfer not possible");
     }
 
+    function batchTransfer(address[] _to, uint256[] _value) external {
```
Nakasar: batchTransfer doesn't make sense I think (an individual holder would never call this method).

Nakasar: Nevermind, it can, why not.

Joachim-Lebrun: yes, it's a request that we had from some issuers, they want to mint the total supply of tokens on their wallet and then proceed to the distribution by using the `transfer` function

Joachim-Lebrun:
we need to wait for @dilipror to write the tests before merging this PR

dilipror:
> we need to wait for @dilipror to write the tests before merging this PR

Done test cases

contracts/registry/IdentityRegistry.sol
```diff
@@ -49,6 +49,12 @@ contract IdentityRegistry is IIdentityRegistry, AgentRole {
         emit IdentityRegistered(_user, _identity);
     }
 
+    function batchRegisterIdentity(address[] calldata _user, Identity[] calldata _identity, uint16[] calldata _country) external {
```
Nakasar: Maybe use plural form for arguments.

```suggestion
    function batchRegisterIdentity(address[] calldata _users, Identity[] calldata _identities, uint16[] calldata _countries) external {
```

contracts/registry/IdentityRegistry.sol
```diff
@@ -49,6 +49,12 @@ contract IdentityRegistry is IIdentityRegistry, AgentRole {
...
+        for (uint256 i = 0; i < _user.length; i++) {
```
Nakasar:
```suggestion
        for (uint256 i = 0; i < _users.length; i++) {
```

contracts/registry/IdentityRegistry.sol
```diff
@@ -49,6 +49,12 @@ contract IdentityRegistry is IIdentityRegistry, AgentRole {
...
+            registerIdentity(_user[i], _identity[i], _country[i]);
```
Nakasar:
```suggestion
            registerIdentity(_users[i], _identities[i], _countries[i]);
```

contracts/token/TransferManager.sol
```diff
@@ -343,27 +355,39 @@ contract TransferManager is Pausable {
...
     /**
      *  Freezes token amount specified for given address.
      *  @param addr The address for which to update frozen tokens
      *  @param amount Amount of Tokens to be frozen
      */
     function freezePartialTokens(address addr, uint256 amount)
-        onlyAgent
-        external
+        public
+		onlyAgent
```
Nakasar: indentation issue here.

test/registry.test.js
```diff
@@ -181,6 +181,15 @@ contract("IdentityRegistry", accounts => {
     const result = await identityRegistry.contains(accounts[3]);
     result.should.equal(false);
   });
+
+  it("Should register identities in batch", async () => {
+	claimHolder3 = await ClaimHolder.new({ from: accounts[3] });
```
Nakasar: Check you identation.

test/registry.test.js
```diff
@@ -181,6 +181,15 @@ contract("IdentityRegistry", accounts => {
...
+    await identityRegistry.batchRegisterIdentity(
+      [accounts[2], accounts[3]],
+      [claimHolder2.address, claimHolder3.address],
+	  [91, 101]
+    );
+  });
```
Nakasar: You should verify that Identities were added by querying them after execution the batch operation, here you only verify that the transaction was a success.

truffle.js
```diff
@@ -53,7 +53,13 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: '0.5.10'
+	  version: '0.5.10',
```
Nakasar: Wrong indentation here?

Joachim-Lebrun:
> Arguments should be put in plural form, and caution with code indentation

done

Nakasar:
I think parameters in other batch methods should also be set to plural form

Joachim-Lebrun:
> Arguments should be put in plural form, and caution with code indentation



> I think parameters in other batch methods should also be set to plural form

done

test/registry.test.js
Joachim-Lebrun: done, for this one and all others

PR #33, opened 2020-02-12
Upgradeable tokens

dilipror:


Joachim-Lebrun:
i really think this feature is unnecessary, we already have all the functions needed to do a clean transition from one token version to the next one and the changes can also be reflected in the onchainID of the token, i think this `upgradeable` feature is something that can be useful with ERC20 but not with TREX tokens where the owner/agent already has all the permissions to freeze the deprecated version of the token, mint the corresponding amount on the new token, burn the old tokens and even transfer the smart contracts of compliance, identity registry, trusted issuers registry and claim topics registry on the next token contract.

PR #39, opened 2020-02-17
Recovery process fix

Joachim-Lebrun:
The recovery function was calling `removeKey` on the onchainID of the investor. In order to remove a management key from an onchainID the contract of the token was supposed to have a management key on the onchainID itself which is against everything we want to achieve with onchainID...


Joachim-Lebrun:
on the contracts deployed before this bugfix, we will have to be sure that the lost wallet management key is removed from the onchainID BEFORE calling the recovery function, otherwise the recovery function will fail

Nakasar:
Why do we event need to test this?

```solidity
bytes32 _key = keccak256(abi.encode(msg.sender));

if (_onchainID.keyHasPurpose(_key, 1)) {
```

https://github.com/TokenySolutions/T-REX/pull/39/files#diff-7d97aa2a3edbb863ef5e5c6a47a7a22bR466-R469

Joachim-Lebrun:
> Why do we event need to test this?
> 
> ```solidity
> bytes32 _key = keccak256(abi.encode(msg.sender));
> 
> if (_onchainID.keyHasPurpose(_key, 1)) {
> ```
> 
> https://github.com/TokenySolutions/T-REX/pull/39/files#diff-7d97aa2a3edbb863ef5e5c6a47a7a22bR466-R469

in case we don't have the management key we need to issue a new onchainID for the investor who lost his wallet, cause we wouldn't be able to add a management key for his new wallet on the onchainID, so i think it's good to check this, even if, imo, we should proceed to the management key change before calling the recovery function

Nakasar:
> imo, we should proceed to the management key change before calling the recovery function.

Yes we should. This will also probably never work otherwise, the wallet that deploy identities is certainly not the same as the one used to own or operate tokens.

Joachim-Lebrun:
> > imo, we should proceed to the management key change before calling the recovery function.
> 
> Yes we should. This will also probably never work otherwise, the wallet that deploy identities is certainly not the same as the one used to own or operate tokens.

indeed, therefore, what we should check is that the new wallet has already a management key on the onchainID, i will make the update

contracts/token/Token.sol
```diff
@@ -4,13 +4,29 @@ import "./TransferManager.sol";
...
+        string memory _version,
+        address _onchainID)
+    public TransferManager(_identityRegistry, _compliance) {
+        name = _name;
+        symbol = _symbol;
+        decimals = _decimals;
+        version = _version;
+        onchainID = _onchainID;
+
+        emit NewTrexTokenCreated(name, symbol, decimals, version, onchainID);
```
Nakasar: I think you can keep the `UpdatedTokenInformation` here instead of creating a new event with the same params.

PR #40, opened 2020-02-18
Updated contracts to solidity version 0.6.0.

dilipror:
These contracts use beta version for openzeppelin solidity contracts now.
Onchain id still needs to be updated to compile and run these contracts.


contracts/token/MintableAndBurnable.sol
```diff
@@ -1,8 +1,8 @@
-pragma solidity ^0.5.10;
+pragma solidity ^0.6.0;
 
 import "./TransferManager.sol";
 
-
+abstract 
 contract MintableAndBurnable is TransferManager {
```
Nakasar: Maybe `interface` instead of `abstract`?

Nakasar: Abstract is more logic, indeed.

Joachim-Lebrun:
> i modified a lot of things in the PR #39 --> please use the branch `recovery-process-fix` instead of `master` as a base for the version update

i just created a PR #41 for that, it will be easier to follow, i took the relevant changes that @dilipror made on PR #40

contracts/token/MintableAndBurnable.sol
Joachim-Lebrun: `MintableAndBurnable` contract doesn't exist anymore, functions were moved to `TransferManager`

contracts/token/Token.sol
```diff
@@ -1,27 +1,40 @@
...
-    address public onchainID = 0x0000000000000000000000000000000000000000;
+    string private tokenName = "TREXDINO";
+    string private tokenSymbol = "TREX";
+    string private tokenVersion = "1.2";
+    uint8 public override decimals = 0;
+    address public override onchainID = 0x0000000000000000000000000000000000000000;
 
     constructor(address _identityRegistry, address _compliance) public TransferManager(_identityRegistry, _compliance) {}
 
+    function name() public view override returns (string memory) {
```
Joachim-Lebrun: i don't see the point of these getters, why not keeping the variables public ?

dilipror: Keeping variables public will make the return types different from getter functions and thus will not allow it to override the IToken state variables

Joachim-Lebrun:
i think we can close this one as the PR #41 is working and up to date with `master` branch

PR #42, opened 2020-02-20
:version:() update TREX to solidity version 0.6.0

Myna65:


contracts/token/Token.sol
```diff
@@ -19,25 +20,75 @@ contract Token is IToken, TransferManager {
...
+    function version() public override view returns (string memory){
+        return tokenVersion;
+    }
+
+    /**
+     * @dev Sets the values for `tokenName`, `tokenSymbol`, `tokenDecimals`,
+     * `tokenVersion` and `tokenOnchainID`
+     */
+
+    function setTokenInformation(string calldata _name, string calldata _symbol, uint8 _decimals, string calldata _version, address _onchainID) external override onlyOwner {
```
Aboudjem: I should be possible to set only one parameter,
setName, setSymbol, etc.

contracts/registry/IdentityRegistry.sol
```diff
@@ -12,12 +12,12 @@ import "@onchain-id/solidity/contracts/Identity.sol";
 
 contract IdentityRegistry is IIdentityRegistry, AgentRole {
     // mapping between a user address and the corresponding identity contract
-    mapping(address => Identity) public identity;
+    mapping(address => Identity) private identity;
```
Aboudjem: Could it be possible to add a mapping between Identity and Address

Myna65: No, because an identity can be linked to multiple wallets

Myna65: At least, it was the original idea

Joachim-Lebrun: we put the addresses in a claim on the onchainID, we can retrieve it from there easily

contracts/token/Token.sol
Myna65: Currently, we have a size problem because smart contracts are too big. It can be interesting once the size will be reduced

contracts/registry/IIdentityRegistry.sol
```diff
@@ -3,8 +3,7 @@ pragma solidity ^0.6.0;
 import "../registry/ITrustedIssuersRegistry.sol";
 import "../registry/IClaimTopicsRegistry.sol";
 
-import "@onchain-id/solidity/contracts/ClaimIssuer.sol";
-import "@onchain-id/solidity/contracts/Identity.sol";
+import "@onchain-id/solidity/contracts/IClaimIssuer.sol";
```
Joachim-Lebrun: `IClaimIssuer` is still importing `Identity` instead of `IIdentity` on the repo of onchainID, we will need to change all the calls to `Identity` into `IIdentity` in the TREX files once it's updated

Myna65:
@Aboudjem All your remarks are interesting but are out of the scope of this PR. Can you open issues with them?

PR #45, opened 2020-02-24
♻ forcedTransfer on partial freeze

Joachim-Lebrun:
forcedTransfer was not possible between partially frozen addresses, which should be the case if we don't want the agent to do 2 transactions (unfreezePartialTokens + forcedTransfer) in the case the amount of free tokens is lower than the amount to transfer, the goal of this PR is to solve this minor issue


Nakasar:
Don't forget to increase the package.json version

Joachim-Lebrun:
> Don't forget to increase the package.json version

done

PR #47, opened 2020-02-24
unbind compliance contracts from token contracts

Joachim-Lebrun:
compliance variables and functions were in the Token smart contracts, it cannot be like this cause with every single custom compliance requirement we should change the TREX contracts, we don't want this at all


contracts/token/TransferManager.sol
```diff
@@ -138,13 +107,33 @@ contract TransferManager is Pausable, ERC20 {
...
+
+    /**
+     * @dev Called by an agent to unpause, returns to normal state.
+     */
+    function unpause() public onlyAgent whenPaused {
+        _paused = false;
+        emit UnPaused(msg.sender);
+    }
+
+    function getIdentityRegistry() public view returns (IIdentityRegistry) {
```
Nakasar: Maybe rename it as a getter `identityRegistry()`? Otherwise you don't event need it, as the state variable is declared as public here https://github.com/TokenySolutions/T-REX/pull/47/files#diff-7d97aa2a3edbb863ef5e5c6a47a7a22bR20

test/tokenTransfer.test.js
```diff
@@ -14,7 +14,7 @@ const ClaimHolder = artifacts.require('@onchain-id/solidity/contracts/Identity.s
 const IssuerIdentity = artifacts.require('@onchain-id/solidity/contracts/ClaimIssuer.sol');
 const Token = artifacts.require('../contracts/token/Token.sol');
 const Compliance = artifacts.require('../contracts/compliance/DefaultCompliance.sol');
-const LimitCompliance = artifacts.require('../contracts/compliance/LimitHolder.sol');
+// const LimitCompliance = artifacts.require('../contracts/compliance/LimitHolder.sol');
```
Nakasar:
```suggestion
```

test/tokenTransfer.test.js
```diff
@@ -23,7 +23,7 @@ contract('Token', accounts => {
   let claimIssuerContract;
   let token;
   let defaultCompliance;
-  let limitCompliance;
+  //  let limitCompliance;
```
Nakasar:
```suggestion
```

contracts/token/TransferManager.sol
```diff
@@ -264,7 +251,7 @@ contract TransferManager is Pausable, ERC20 {
         require(identityRegistry.isVerified(_to), "Identity is not verified.");
 
         _mint(_to, _amount);
-        updateShareholders(_to);
+        compliance.created(_to, _amount);
```
dilipror: Compliance should be checked before minting to revert in cases where compliance will reject the transfers

Joachim-Lebrun: yes, it should be like this, but rather in `mint()` than in `_mint()` in my opinion

contracts/token/TransferManager.sol
Joachim-Lebrun: you need it as the state varable is not accessible from interfaces and as we only import interfaces in external contracts, we could change the getters into `identityRegistry()` and `compliance()` though but i don't really think it's a necessity

Joachim-Lebrun: and the state variables have been updated to `private` in the next PR

contracts/token/TransferManager.sol
dilipror: _mint() is a standard function we have derived from ERC20 like _transfer and other methods. Imo we should include TREX specific methods here.

contracts/token/TransferManager.sol
Nakasar: Me are more used to have getter in the form "identityRegistry" instead of "getIdentityRegistry" (take example on ERCx interfaces, you don't have getBalanceOf() or getTokenSupply(), you have balanceOf() and tokenSupply())

PR #49, opened 2020-02-26
new contracts : AgentManager and OwnerManager

Joachim-Lebrun:
those smart contracts are intended to be used to increase granularity of the agent/owner roles, as the roles described at token level could be too permissive and as the role's permissions are hardcoded in the token smart contracts and not modifiable post-deployment, these external smart contracts could be used easily to change the permissions of a role, and a custom version of this smart contract could also be created in case there is a need for even more granularity in the role's permissions. By setting this smart contract as agent/owner on the smart contracts it allows to get more granularity on the permissions linked to the different roles and it also allows to link these roles to the onchainID of the agent/owner which prevent him from losing access to the functions if he loose his private key (as it is recommanded to have several management keys on the onchainID)

### DIFF
--- a/contracts/compliance/LimitHolder.sol
+++ b/contracts/compliance/LimitHolder.sol
@@ -115,11 +115,8 @@ contract LimitHolder is ICompliance, AgentRole {
 
     function created(address _to, uint256 _value) public override onlyAgent returns (bool) {
         require(_value > 0, "No token created");
-        if (holderCount() < holderLimit) {
-            updateShareholders(_to);
-            return true;
-        }
-        return false;
+        updateShareholders(_to);
+        return true;
     }
 
     function destroyed(address _from, uint256 _value) public override onlyAgent returns (bool) {
--- a/contracts/compliance/LimitHolder.sol
+++ b/contracts/compliance/LimitHolder.sol
@@ -8,7 +8,7 @@ import "../registry/IIdentityRegistry.sol";
 contract LimitHolder is ICompliance, AgentRole {
     IToken public token;
     uint public holderLimit;
-    IIdentityRegistry private identityRegistry = token.getIdentityRegistry();
+    IIdentityRegistry private identityRegistry;
     mapping(address => uint256) private holderIndices;
     mapping(uint16 => uint256) private countryShareHolders;
     address[] private shareholders;
@@ -16,6 +16,7 @@ contract LimitHolder is ICompliance, AgentRole {
     constructor (address _token, uint _holderLimit) public {
         token = IToken(_token);
         holderLimit = _holderLimit;
+		identityRegistry = token.getIdentityRegistry();
     }
 
 
@@ -61,7 +62,7 @@ contract LimitHolder is ICompliance, AgentRole {
      *  @dev see https://ethereum.stackexchange.com/a/39311
      */
     function pruneShareholders(address addr, uint256 value) internal {
-        uint256 balance = token.balanceOf(addr) - value;
+        uint256 balance = token.balanceOf(addr);
         if (balance > 0) {
             return;
         }
@@ -97,7 +98,10 @@ contract LimitHolder is ICompliance, AgentRole {
     * @param _value The amount of tokens involved in the transfer
     */
     function canTransfer(address _from, address _to, uint256 _value) public override view returns (bool) {
-        if (holderCount() < holderLimit) {
+        if (holderIndices[_to] != 0) {
+			return true;
+		}
+		if (holderCount() < holderLimit) {
             return true;
         }
         return false;
@@ -111,8 +115,11 @@ contract LimitHolder is ICompliance, AgentRole {
 
     function created(address _to, uint256 _value) public override onlyAgent returns (bool) {
         require(_value > 0, "No token created");
-        updateShareholders(_to);
-        return true;
+        if (holderCount() < holderLimit) {
+        	updateShareholders(_to);
+        	return true;
+        }
+		return false;
     }
 
     function destroyed(address _from, uint256 _value) public override onlyAgent returns (bool) {
--- a/contracts/compliance/LimitHolder.sol
+++ b/contracts/compliance/LimitHolder.sol
@@ -1,3 +1,26 @@
+/**
+ *     NOTICE
+ *
+ *     The T-REX software is licensed under a proprietary license or the GPL v.3.
+ *     If you choose to receive it under the GPL v.3 license, the following applies:
+ *     T-REX is a suite of smart contracts developed by Tokeny to manage and transfer financial assets on the ethereum blockchain
+ *
+ *     Copyright (C) 2019, Tokeny sàrl.
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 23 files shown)


Joachim-Lebrun:
@dilipror could you please write tests to get back to 100% coverage? thanks :)

contracts/roles/AgentManager.sol
```diff
@@ -0,0 +1,83 @@
...
+import "../registry/IIdentityRegistry.sol";
+import "./AgentRoles.sol";
+import "@onchain-id/solidity/contracts/IIdentity.sol";
+
+contract AgentManager is Ownable, AgentRoles {
+
+    IToken public token;
+    IIdentityRegistry public identityRegistry;
+
+
```
Nakasar:
```suggestion
```

contracts/roles/AgentRoles.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    Roles.Role private _supplyModifiers;
+    Roles.Role private _freezers;
+    Roles.Role private _transferManagers;
+    Roles.Role private _recoveryAgents;
+    Roles.Role private _complianceAgents;
+    Roles.Role private _whiteListManagers;
+
+
+    // SupplyModifier Role _supplyModifiers
```
Nakasar:
```suggestion
```

contracts/roles/AgentRoles.sol
```diff
@@ -0,0 +1,117 @@
...
+    event RoleRemoved(address indexed account);
+
+    Roles.Role private _supplyModifiers;
+    Roles.Role private _freezers;
+    Roles.Role private _transferManagers;
+    Roles.Role private _recoveryAgents;
+    Roles.Role private _complianceAgents;
+    Roles.Role private _whiteListManagers;
+
+
```
Nakasar:
```suggestion
```

contracts/roles/AgentRoles.sol
```diff
@@ -0,0 +1,117 @@
...
+    // SupplyModifier Role _supplyModifiers
+
```
Nakasar:
```suggestion
```

contracts/roles/AgentRoles.sol
```diff
@@ -0,0 +1,117 @@
...
+    function addWhiteListManager(address account) public onlyOwner {
+        _whiteListManagers.add(account);
+        emit RoleAdded(account);
+    }
+
+    function removeWhiteListManager(address account) public onlyOwner {
+        _whiteListManagers.remove(account);
+        emit RoleRemoved(account);
+    }
+
```
Nakasar:
```suggestion
```

contracts/roles/AgentRoles.sol
```diff
@@ -0,0 +1,117 @@
...
+    Roles.Role private _whiteListManagers;
+
+
+    // SupplyModifier Role _supplyModifiers
+
+    function isSupplyModifier(address account) public view returns (bool) {
+        return _supplyModifiers.has(account);
+    }
+
+    function addSupplyModifier(address account) public onlyOwner {
```
Nakasar: Maybe we should add a role authorized to update role members instead of leaving it to the Owner only.

contracts/roles/AgentRoles.sol
```diff
@@ -15,20 +15,41 @@ contract AgentRoles is Ownable {
...
+        _;
+    }
+
+    // AgentAdmin Role _agentAdmin
+
+    function isAgentAdmin(address account) public view returns (bool) {
+        return _agentAdmin.has(account);
+    }
+
+    function addAgentAdmin(address account) public onlyOwner {
```
Nakasar: Maybe even:
```suggestion
    function addAgentAdmin(address account) public onlyAdmin {
```
?

Myna65: Why not

Joachim-Lebrun: yes, true, modif is done

contracts/roles/AgentManager.sol
```diff
@@ -0,0 +1,112 @@
...
+import "@onchain-id/solidity/contracts/IIdentity.sol";
+
+contract AgentManager is AgentRoles {
+
+    IToken public token;
+    IIdentityRegistry public identityRegistry;
+
+    constructor (address _token) public {
+        token = IToken(_token);
+        identityRegistry = token.getIdentityRegistry();
```
Nakasar: I think we can remove the state variable for identity registry and retrieve it each time from the token contract.

contracts/roles/AgentManager.sol
```diff
@@ -0,0 +1,112 @@
+pragma solidity ^0.6.0;
+
+import "../token/IToken.sol";
+import "../registry/IIdentityRegistry.sol";
+import "./AgentRoles.sol";
+import "@onchain-id/solidity/contracts/IIdentity.sol";
+
+contract AgentManager is AgentRoles {
```
Nakasar: Maybe we can add those permission system to the TrustedIssuerRegistry as well.

Joachim-Lebrun: yes, indeed, but it's not for the agent role anymore as TrustedIssuerRegistry is only `Ownable` i was planning to write an `OwnerManager` contract for all the `onlyOwner` functions

Nakasar: Shouldn't the TrustedIssuerRegistry be managed by an Agent as well?

Joachim-Lebrun: no, it's the owner who decides who are the trusted issuers, the same way as he decides who are the agents :)

test/agentManager.test.js
```diff
@@ -131,229 +126,265 @@ contract('Agent Manager', accounts => {
...
         (await agentManager.isWhiteListManager(user1Contract.address)).should.be.equal(true);
         await identityRegistry.addAgent(agentManager.address, { from: tokeny });
-        await agentManager.callDeleteIdentity(user2, user1Contract.address, { from: user1 });
+	    await agentManager.callDeleteIdentity(user2, user1Contract.address, { from: user1 });
+	    const result = await identityRegistry.contains(user2);
+	    result.should.equal(false);
+	});
+	
+	it('Should remove whitelist manager from the role manager.', async () => {
+        await agentManager.addWhiteListManager(user1Contract.address, { from: admin });
```
Joachim-Lebrun:
```suggestion
        await agentManager.addWhiteListManager(user1Contract.address, { from: admin });
        (await agentManager.isWhiteListManager(user1Contract.address)).should.be.equal(true);
```

contracts/compliance/LimitHolder.sol
```diff
@@ -111,8 +115,11 @@ contract LimitHolder is ICompliance, AgentRole {
 
     function created(address _to, uint256 _value) public override onlyAgent returns (bool) {
         require(_value > 0, "No token created");
-        updateShareholders(_to);
-        return true;
+        if (holderCount() < holderLimit) {
```
Joachim-Lebrun: that's not the role of the `created()` function to check if the transaction can be processed, you have to use the `canTransfer()` for that. `created()` `destroyed()` and `transferred()` are called only to update the shareholders list in this case, or to do some other stuffs in different compliance contracts post-transaction. in the case of a minting process you should just use the `canTransfer()` function with `_msgSender()` as a `_from`

contracts/token/Token.sol
```diff
@@ -526,9 +526,9 @@ contract Token is IToken, Context, AgentRole {
      */
     function mint(address _to, uint256 _amount) public override onlyAgent {
         require(identityRegistry.isVerified(_to), "Identity is not verified.");
```
Joachim-Lebrun: use `canTransfer()` in the `require` instead of using a `if` on `created()`

contracts/registry/TrustedIssuersRegistry.sol
```diff
@@ -182,12 +205,27 @@ contract TrustedIssuersRegistry is ITrustedIssuersRegistry, Ownable {
         for (i = 0; i < claimTopicsLength; i++) {
             trustedIssuerClaimTopics[index][i] = claimTopics[i];
         }
+        delete trustedIssuer[address(trustedIssuers[index])];
```
Joachim-Lebrun: there was a bug on this function as it was not updating `trustedIssuer`. the function `isTrustedIssuer` was not returning the good result after issuer update

PR #51, opened 2020-02-27
External owner manager

Joachim-Lebrun:
this smart contract is intended to be used to increase granularity of the owner role, as the owner role could be too permissive and as the role's permissions are hardcoded in the token smart contracts and not modifiable post-deployment, this external smart contract could be used easily to change the permissions of a role, and a custom version of this smart contract could also be created in case there is a need for even more granularity in the role's permissions. By setting this smart contract as owner on the Token and all the other smart contracts of TREX it allows to get more granularity on the permissions linked to the owner role and it also allows to link this role to the onchainID of the owner which prevent him from losing access to the functions if he loose his private key (as it is recommanded to have several management keys on the onchainID)

### DIFF
--- a/contracts/roles/OwnerManager.sol
+++ b/contracts/roles/OwnerManager.sol
@@ -1,0 +1,112 @@
+pragma solidity ^0.6.0;
+
+import "../token/IToken.sol";
+import "../registry/IIdentityRegistry.sol";
+import "../registry/ITrustedIssuersRegistry.sol";
+import "../registry/IClaimTopicsRegistry.sol";
+import "../compliance/ICompliance.sol";
+import "./OwnerRoles.sol";
+import "@onchain-id/solidity/contracts/IIdentity.sol";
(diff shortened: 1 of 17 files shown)


contracts/roles/OwnerManager.sol
```diff
@@ -0,0 +1,112 @@
...
+        identityRegistry = token.getIdentityRegistry();
+        topicsRegistry = identityRegistry.getTopicsRegistry();
+        topicsRegistry.transferOwnershipOnClaimTopicsRegistryContract(newOwner);
+    }
+
+    function callTransferOwnershipOnIssuersRegistryContract(address newOwner) external onlyAdmin {
+        identityRegistry = token.getIdentityRegistry();
+        issuersRegistry = identityRegistry.getIssuersRegistry();
+        issuersRegistry.transferOwnershipOnIssuersRegistryContract(newOwner);
+    }
```
Joachim-Lebrun: a function to set agents on the smart contracts that are AgentRole is still missing, i will add it soon :-)

PR #48, opened 2020-02-25
♻() refactor token contract and interface

Joachim-Lebrun:

PR #61, opened 2020-03-29
TREX V3.0.0

Joachim-Lebrun:
* ♻() refactor token contract and interface (#48)

          * ♻() refactor token contract and interface

* new contracts : AgentManager and OwnerManager (#49)

          * ➕() add AgentManager contract

          * ➕() add OwnerManager contract (#51)

          * ➕() add functions to manage Agents and Owners from the interfaces of TREX           
          contracts

          * ✅() add tests on OwnerRoles and OwnerManager

          * 🐛() fix bug on burn function

          * 📝() add license notice on smart contracts

* Docs/natspec (#56)

          * 📝() documentation of contracts

* Docs/natspec (#58)

          * 📝() documentation update + variable names harmonization

          * ✅(Token) freeze status transfer on recovery

* BREAKING CHANGES : refactoring Issuers registry (#59)

          * ♻() refactoring Issuers registry

* ♻() modify token info setters

* 🔖() change TREX version to 3.0.0 after all changes

* ➕() Import ownable and role as they were removed from OZ (#64)


contracts/compliance/ICompliance.sol
```diff
@@ -1,5 +1,83 @@
...
+    *  @dev function called whenever tokens are created
+    *  on a wallet
+    *  this function can update state variables in the compliance contract
+    *  these state variables being used by `canTransfer` to decide if a transfer
+    *  is compliant or not depending on the values stored in these state variables and on
+    *  the parameters of the compliance smart contract
+    *  @param _to The address of the receiver
+    *  @param _amount The amount of tokens involved in the transfer
+    */
+    function created(address _to, uint256 _amount) external returns (bool);
```
Nakasar: Maybe we should remove `created` and `destroyed` methods and use `transferred` instead with `address(0)` for `from` and `to`?
Maybe it's better to keep destroyed and created, but in that case maybe we could call them `onMint` and `onBurn`?

contracts/compliance/ICompliance.sol
```diff
@@ -1,5 +1,83 @@
...
+    *  from one wallet to another
+    *  this function can update state variables in the compliance contract
+    *  these state variables being used by `canTransfer` to decide if a transfer
+    *  is compliant or not depending on the values stored in these state variables and on
+    *  the parameters of the compliance smart contract
+    *  @param _from The address of the sender
+    *  @param _to The address of the receiver
+    *  @param _amount The amount of tokens involved in the transfer
+    */
+    function transferred(address _from, address _to, uint256 _amount) external returns (bool);
```
Nakasar: This function is purely informative for the compliance contract, I think it shouldn't return a boolean.

contracts/compliance/ICompliance.sol
```diff
@@ -1,5 +1,83 @@
...
+   /**
+    *  @dev function called whenever tokens are destroyed
+    *  this function can update state variables in the compliance contract
+    *  these state variables being used by `canTransfer` to decide if a transfer
+    *  is compliant or not depending on the values stored in these state variables and on
+    *  the parameters of the compliance smart contract
+    *  @param _from The address of the receiver
+    *  @param _amount The amount of tokens involved in the transfer
+    */
+    function destroyed(address _from, uint256 _amount) external returns (bool);
```
Nakasar: This function is purely informative for the compliance contract, I think it shouldn't return a boolean.

contracts/compliance/ICompliance.sol
Nakasar: This function is purely informative for the compliance contract, I think it shouldn't return a boolean.

Joachim-Lebrun: you are right, i will change all of them to return nothing, on the naming i pretty like the destroyed and created functions though ...

Joachim-Lebrun: solved here https://github.com/TokenySolutions/T-REX/commit/48f859a6eb7bf734b9cb014b0e1b74cf40453993

contracts/compliance/ICompliance.sol
Joachim-Lebrun: solved here https://github.com/TokenySolutions/T-REX/commit/48f859a6eb7bf734b9cb014b0e1b74cf40453993

contracts/compliance/ICompliance.sol
Joachim-Lebrun: solved here https://github.com/TokenySolutions/T-REX/commit/48f859a6eb7bf734b9cb014b0e1b74cf40453993

Nakasar:
If we don't intend to keep compatibility with v2, maybe we can remove underscores from all event parameters? It's a bit weird and unusual I think.

```solidity
function deleteIdentity(address _userAddress) external;
```

Into

```solidity
function deleteIdentity(address userAddress) external;
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tokenysolutions/t-rex",
-  "version": "2.5.1",
+  "version": "3.0.0-beta1",
```
Nakasar, line 3:
```suggestion
  "version": "3.0.0-beta2",
```

Nakasar:
Deployed with beta tag as `@tokenysolutions/t-rex@3.0.0-beta2`.

Nakasar:
Published with beta tag as `@tokenysolutions/t-rex@3.0.0-beta3`.

.github/workflows/publish-release.yml
```diff
@@ -38,8 +39,8 @@ jobs:
         with:
           node-version: 12
           registry-url: https://npm.pkg.github.com/
-          scope: '@tokenysolutions'
-      - run: npm ci
-      - run: npm publish
+          scope: '@onchain-id'
```
Aboudjem, line 42: Should this not be `@TokenySolutions` instead?

PR #64, opened 2020-03-31
➕() Import ownable and role as they were removed from OZ

Nakasar:
OppenZeppelin overrided their 3.0.0-beta.0 version and removed the Roles.sol and moved the Ownable.sol.

The implementation of Roles is no longer compatible. I imported the old version in our codebase, as well as the old Ownable.

### DIFF
--- a/contracts/compliance/DefaultCompliance.sol
+++ b/contracts/compliance/DefaultCompliance.sol
@@ -24,7 +24,8 @@
 pragma solidity ^0.6.0;
 
 import "./ICompliance.sol";
-import "openzeppelin-solidity/contracts/access/Ownable.sol";
+
+import "../Roles/Ownable.sol";
 
 contract DefaultCompliance is ICompliance, Ownable {
 
--- a/contracts/compliance/DefaultCompliance.sol
+++ b/contracts/compliance/DefaultCompliance.sol
@@ -25,7 +25,7 @@ pragma solidity ^0.6.0;
 
 import "./ICompliance.sol";
 
-import "../Roles/Ownable.sol";
+import "../roles/Ownable.sol";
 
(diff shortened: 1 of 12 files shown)


contracts/registry/ClaimTopicsRegistry.sol
```diff
@@ -24,7 +24,7 @@
 pragma solidity ^0.6.0;
 
 import "../registry/IClaimTopicsRegistry.sol";
-import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "openzeppelin-solidity/contracts/access/Ownable.sol";
```
Joachim-Lebrun:
```suggestion
import "../roles/Ownable.sol";
```

contracts/compliance/DefaultCompliance.sol
```diff
@@ -24,7 +24,7 @@
 pragma solidity ^0.6.0;
 
 import "./ICompliance.sol";
-import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "openzeppelin-solidity/contracts/access/Ownable.sol";
```
Joachim-Lebrun:
```suggestion
import "../roles/Ownable.sol";
```

contracts/compliance/DefaultCompliance.sol
```diff
@@ -24,7 +24,8 @@
 pragma solidity ^0.6.0;
 
 import "./ICompliance.sol";
-import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+
+import "../Roles/Ownable.sol";
```
Joachim-Lebrun:
```suggestion
import "../roles/Ownable.sol";
```

PR #66, opened 2020-04-02
➕(Destroyable) Added destroyable to contracts + tests

Aboudjem:
Destroyable

Added Destroyable on:

- Token
- TrustedIssuers
- Identity
- IdentityStorage
- Compliamce
- ClaimTopics

Added tests

### DIFF
--- a/test/destroyable.test.js
+++ b/test/destroyable.test.js
@@ -1,0 +1,96 @@
+require('chai')
+  .use(require('chai-as-promised'))
+  .should();
+
+const ClaimTopicsRegistry = artifacts.require('../contracts/registry/ClaimTopicsRegistry.sol');
+const IdentityRegistry = artifacts.require('../contracts/registry/IdentityRegistry.sol');
+const TrustedIssuersRegistry = artifacts.require('../contracts/registry/TrustedIssuersRegistry.sol');
+const ClaimHolder = artifacts.require('@onchain-id/solidity/contracts/Identity.sol');
+const Token = artifacts.require('../contracts/token/Token.sol');
+const Compliance = artifacts.require('../contracts/compliance/DefaultCompliance.sol');
+const IdentityRegistryStorage = artifacts.require('../contracts/registry/IdentityRegistryStorage.sol');
+
+contract('Destruct', accounts => {
+  let claimTopicsRegistry;
+  let identityRegistry;
+  let trustedIssuersRegistry;
+  let identityRegistryStorage;
+  let token;
+  let defaultCompliance;
+  let tokenName;
+  let tokenSymbol;
+  let tokenDecimals;
+  let tokenVersion;
+  let tokenOnchainID;
+
+  const tokeny = accounts[0];
+
+  beforeEach(async () => {
+    // Tokeny deploying token
+
+    claimTopicsRegistry = await ClaimTopicsRegistry.new({ from: tokeny });
+    trustedIssuersRegistry = await TrustedIssuersRegistry.new({ from: tokeny });
+    defaultCompliance = await Compliance.new({ from: tokeny });
+    identityRegistryStorage = await IdentityRegistryStorage.new({ from: tokeny });
+    identityRegistry = await IdentityRegistry.new(trustedIssuersRegistry.address, claimTopicsRegistry.address, identityRegistryStorage.address, {
+      from: tokeny,
+    });
+    tokenOnchainID = await ClaimHolder.new({ from: tokeny });
+    tokenName = 'TREXDINO';
+    tokenSymbol = 'TREX';
+    tokenDecimals = '0';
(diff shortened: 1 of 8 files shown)


test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+      { from: tokeny },
+    );
+  });
+
+  it('name returns the name of the token', async () => {
+    const name1 = await token.name().should.be.fulfilled;
+    name1.toString().should.equal('TREXDINO');
+  });
+
+  it('should destroy the token contract', async () => {
```
Joachim-Lebrun, line 60:
```suggestion
  it('should destroy the token contract', async () => {
    const name1 = await token.name().should.be.fulfilled;
    name1.toString().should.equal('TREXDINO');
```

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+      tokenName,
+      tokenSymbol,
+      tokenDecimals,
+      tokenVersion,
+      tokenOnchainID.address,
+      { from: tokeny },
+    );
+  });
+
+  it('name returns the name of the token', async () => {
```
Joachim-Lebrun, line 55:
```suggestion
```

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const name1 = await token.name().should.be.fulfilled;
```
Joachim-Lebrun, line 56:
```suggestion
```

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    name1.toString().should.equal('TREXDINO');
```
Joachim-Lebrun, line 57:
```suggestion
```

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+  });
```
Joachim-Lebrun, line 58:
```suggestion
```

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+  it('should destroy the token contract', async () => {
+    await token.setDestroyAuthorization(true);
+    await token.destroyContract();
+    await token.name().should.be.rejected;
+  });
+
+  it('should destroy the Identity Registry contract and make it impossible to call', async () => {
+    await identityRegistry.setDestroyAuthorization(true);
```
Joachim-Lebrun, line 67: better to not use the 0 address, it can be rejected at different layers by the `require` clauses, it would be better if you make a test that works on the identity registry first and then test that it fails when it's destroyed

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    await identityRegistry.destroyContract();
+    await identityRegistry.isVerified('0x0000000000000000000000000000000000000000').should.be.rejected;
+  });
+
+  it('should destroy the Global Identity Registry contract and make it impossible to call', async () => {
+    await identityRegistryStorage.setDestroyAuthorization(true);
```
Joachim-Lebrun, line 73: same as for identity registry, make it work first and then try with the same address, no 0 address

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    await identityRegistryStorage.destroyContract();
+    await identityRegistryStorage.storedIdentity('0x0000000000000000000000000000000000000000').should.be.rejected;
+  });
+
+  it('should destroy the ClaimTopics Registry contract and make it impossible to call', async () => {
+    await claimTopicsRegistry.setDestroyAuthorization(true);
```
Joachim-Lebrun, line 79: same as for identity registry, make it work first and then try with the same function call

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    await claimTopicsRegistry.destroyContract();
+    await claimTopicsRegistry.getClaimTopics().should.be.rejected;
+  });
+
+  it('should destroy the TrustedIssuers Registry contract and make it impossible to call', async () => {
+    await trustedIssuersRegistry.setDestroyAuthorization(true);
```
Joachim-Lebrun, line 85: same as for identity registry, make it work first and then try with the same address, no 0 address

test/destroyable.test.js
```diff
@@ -0,0 +1,96 @@
...
+    await trustedIssuersRegistry.destroyContract();
+    await trustedIssuersRegistry.isTrustedIssuer('0x0000000000000000000000000000000000000000').should.be.rejected;
+  });
+
+  it('should destroy the Compliance and make it impossible to call', async () => {
+    await defaultCompliance.setDestroyAuthorization(true);
```
Joachim-Lebrun, line 91: same as for identity registry, make it work first and then try with the same address, no 0 address

PR #67, opened 2020-04-02
Refactoring identity

Aboudjem:
Changed double mapping into one, identity is now a struct containing the IdentityContract and InvestorCountry


Joachim-Lebrun:
looks better indeed  👍

Aboudjem:
> looks better indeed 👍

thanks ! I think we can merge  country and Identity getters into one getter (returning struct)

PR #69, opened 2020-04-07
Update Contracts Interface

Aboudjem:
Added IdentityRegistryStorage to the interfaces


.github/workflows/push_checking.yml
```diff
@@ -50,5 +50,8 @@ jobs:
 
         run: yarn install --frozen-lockfile
 
+      - name: check compilers
+        run: yarn truffle compile --list
+
```
Nakasar, line 55:
```suggestion
```

PR #70, opened 2020-04-08
Version hardcoded

Aboudjem:
- Hardcoded "**Version**" Into the Constructor
- Updated Tests


claimtopicsregistry.sol
```diff
@@ -0,0 +1,260 @@
+
+// File: contracts/registry/IClaimTopicsRegistry.sol
```
Myna65, line 2: What is this file?

PR #71, opened 2020-04-08
Remove version

Aboudjem:
Removed version from constructor, hardcoded version "3.0"

Updated tests


contracts/token/Token.sol
```diff
@@ -81,14 +79,13 @@ contract Token is IToken, AgentRole {
...
         uint8 _decimals,
-        string memory _version,
         address _onchainID
         )
     public {
         tokenName = _name;
         tokenSymbol = _symbol;
         tokenDecimals = _decimals;
-        tokenVersion = _version;
+        tokenVersion = "3.0";
```
Nakasar, line 88: Maybe include hotfix version as well?

```suggestion
        tokenVersion = "3.0.0";
```

Nakasar:
I think we can directly declare the variable as constant and remove it from the constructor.

Nakasar:
This is also a breaking change to mirror in our SDK/documentation.

Joachim-Lebrun:
i agree that, in this case it is better to put it as a constant directly and remove it prom the constructor, logically the event associated, `UpdatedTokenInformation(tokenName, tokenSymbol, tokenDecimals, tokenVersion, tokenOnchainID)` should not include the version in this case but it would break the tracking done on thegraph

Aboudjem:
> i agree that, in this case it is better to put it as a constant directly and remove it from the constructor, logically the event associated, `UpdatedTokenInformation(tokenName, tokenSymbol, tokenDecimals, tokenVersion, tokenOnchainID)` should not include the version in this case but it would break the tracking done on thegraph

Yes that's why I let the UpdatedTokenInformation, I think it's better to let it like that so we don't have to do changes on Graph and everytime a Token is deployed we have all the information at once

Myna65:
> i agree that, in this case it is better to put it as a constant directly and remove it from the constructor, logically the event associated, `UpdatedTokenInformation(tokenName, tokenSymbol, tokenDecimals, tokenVersion, tokenOnchainID)` should not include the version in this case but it would break the tracking done on thegraph

Or we can keep it in the event and just hardcode it there too

Nakasar:
Increase package-lock version to "3.0.0-betaX"

contracts/token/Token.sol
```diff
@@ -69,8 +69,6 @@ contract Token is IToken, AgentRole {
     *  @param _name the name of the token
     *  @param _symbol the symbol of the token
     *  @param _decimals the decimals of the token
-    *  @param _version the version of the token, current version is 3.0
-    *  @param _onchainID the address of the onchainID of the token
```
Joachim-Lebrun, line 73: why did you remove this?

Joachim-Lebrun: _onchainID is still a parameter

Aboudjem: > _onchainID is still a parameter

Didn't notice that I have deleted this, fixed :)

PR #72, opened 2020-04-16
Proposal/refactoring token contract

Aboudjem:
### Refactoring Token Contract

- Removed setSymbol from Token
- Removed setName from Token
- Removed tests on both functions
- Removed CallSetName from OwnerManager
- Removed CallSetSymbol from OwnerManager
- Added `Require `on empty strings making it impossible to set an empty `Name `or `Symbol `

It should be impossible to set an empty **Name** or **Symbol**, and it should be impossible to modify **Name** or **Symbol** as it could lead to _undefined behavior_ as previously encountered with **Etherscan**

We should consider removing **TokenOnchainID** from the constructor, as it is unnecessary to Add a **zeroAddress** and set it afterward, but like we said it could break many services like Graph

### DIFF
--- a/contracts/roles/OwnerManager.sol
+++ b/contracts/roles/OwnerManager.sol
@@ -73,32 +73,6 @@ contract OwnerManager is OwnerRoles {
     }
 
    /**
-    *  @dev calls the `setName` function on the token contract
-    *  OwnerManager has to be set as owner on the token smart contract to process this function
-    *  See {IToken-setName}.
-    *  Requires that `_onchainID` is set as TokenInfoManager on the OwnerManager contract
-    *  Requires that msg.sender is a MANAGEMENT KEY on `_onchainID`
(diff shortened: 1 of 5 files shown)


Joachim-Lebrun:
i don't understand the purpose of this PR ...

Joachim-Lebrun:
> We should consider removing **TokenOnchainID** from the constructor, as it is unnecessary to Add a **zeroAddress** and set it afterward, but like we said it could break many services like Graph

Where did you see that we are supposed to add the zero address in the constructor?? you are supposed to set an onchainID address at the token deployment

Myna65:
We are developping smart contrats not workaround for etherscan bugs. if we want to update the name we should be able to do so. If etherscan does not update it, we should just contact them.

This is a breaking change and should be critical. Unfortunately this is not the case

PR #75, opened 2020-05-27
Update Compliance methods order

Aboudjem:
Changed order of the compliance method (Created, Destroyed) on mint and burn.
As Compliance's methods should be called before mint or burn (like on transfers methods)


Joachim-Lebrun:
i don't agree with this. the `canTransfer()` method has to be called pre-transfer, yes, but the `created()` `transferred()` and `destroyed()` methods are there only to update the state variables  on the compliance contract, not to ensure the compliance of the methods calling them. If compliance rules have to be applied on a function it has to call `canTransfer()`

Myna65:
This is a breaking change since it affect the behavior of the SCs.
It'll be error prone since the same compliance contract should behave the same on all trex v3 contracts

PR #77, opened 2020-06-05
🐛(Packages) Upgraded all packages, removed unused ethers,removed unus…

Aboudjem:
- Upgraded packages
- Removed Ethers package(not used)
- Removed index.d.ts (not used)


index.d.ts
```diff
@@ -1,17 +0,0 @@
-export namespace contracts {
```
Nakasar, line 1: This is required when using the package in other projects, keep it up to date.

PR #78, opened 2020-06-09
🔖() Set version on Token

Nakasar:


Aboudjem:
You need to update the test too

PR #83, opened 2020-10-19
🐛(IdentityRegistry) fix isVerified function

Joachim-Lebrun:
fixing a bug on isVerified() function on the identity registry

Description of the bug : 
If an onchainID contains several claims of the same claim topic and that this claim topic is required by the token, the isVerified function will return false if the correct claim (issued by the right claim issuer) is not the last one in the array of claim that was fetched by the isVerified function.


Nakasar:
Conflicts resolved in https://github.com/TokenySolutions/T-REX/pull/90

PR #85, opened 2021-01-11
Features/proxy

Aboudjem:


contracts/compliance/LimitHolder.sol
```diff
@@ -173,7 +173,7 @@ contract LimitHolder is ICompliance, Ownable
      *  @return `address` the address of the token holder with the given index.
      */
     function holderAt(uint256 index) external view returns (address) {
-        require(index < shareholders.length, "shareholder doesn't exist");
+        require(index < shareholders.length, 'shareholder doesn\'t exist');
```
Nakasar, line 176: For solidity, I think it's better to use double quotes, isn't it?

contracts/compliance/ICompliance.sol
```diff
@@ -87,66 +86,73 @@ interface ICompliance {
...
-   /**
-    *  @dev function called whenever tokens are destroyed
-    *  this function can update state variables in the compliance contract
-    *  these state variables being used by `canTransfer` to decide if a transfer
-    *  is compliant or not depending on the values stored in these state variables and on
-    *  the parameters of the compliance smart contract
-    *  @param _from The address of the receiver
-    *  @param _amount The amount of tokens involved in the transfer
-    */
+    /**
```
Joachim-Lebrun, line 139: would have been a good idea to lint in another PR, it's very hard to see the real changes, that are related to the proxy addition....

contracts/compliance/LimitHolder.sol
Joachim-Lebrun: that's the most used form indeed, both work but most standard contracts use double quotes

.github/workflows/push_checking.yml
```diff
@@ -18,18 +18,11 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install dependencies
-
-        run: yarn install --frozen-lockfile
```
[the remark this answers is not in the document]
mario-tokeny: Hey, I was going over this and please never never do `npm install` unless you want to get unexpected results.

The lock file (package-lock.json or yarn.lock) has a purpose. And it is that everybody that _installs_ the dependencies gets exactly the same versions. That's what `npm ci` will do. It will install exactly the same versions listed in the lock file. On the other hand, `npm install` will get the latest packages for the direct dependencies and sub-dependencies that meet the requirements. Hence, everybody will have different versions, including the package that was built and released.

The problem arises when due to mismatching versions, someone does not have a bug, while other developer or user has.

.github/workflows/push_checking.yml
```diff
@@ -18,18 +18,11 @@ jobs:
...
+        run: npm install
```
mario-tokeny, line 21:
```suggestion
        run: npm ci
```

The same applies for running the tests. Once a package has been modified in the package-lock and committed, the rest of tasks should use exclusively that version.

.github/workflows/push_checking.yml
```diff
@@ -47,8 +40,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install dependencies
+        run: npm install
```
mario-tokeny, line 43:
```suggestion
        run: npm ci
```

.github/workflows/push_checking.yml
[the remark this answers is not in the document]
Joachim-Lebrun: the best solution in my opinion would be to keep the `package-lock.json` but to replace `npm i` by `npm ci`

Aboudjem:
revert yarn

Nakasar (approved):

package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
```
Joachim-Lebrun, line 16: change version to 3.3.0 in package.json

contracts/token/Storage.sol
```diff
@@ -0,0 +1,29 @@
...
+    mapping(address => uint256) internal _balances;
+    mapping(address => mapping(address => uint256)) internal _allowances;
+    uint256 internal _totalSupply;
+
+    /// Token information
+    string internal tokenName;
+    string internal tokenSymbol;
+    uint8 internal tokenDecimals;
+    address internal tokenOnchainID;
+    string internal constant TOKEN_VERSION = '3.2.0';
```
Joachim-Lebrun, line 16: change version to 3.3.0.

Joachim-Lebrun (changes requested):

Joachim-Lebrun (approved):

PR #125, opened 2022-12-06
Hacken corrections

Joachim-Lebrun:
this PR will contain all corrections required from audit by Hacken

Critical :
- [x] check accessibility of IRS functions to bind and unbind IRS 
- [x] fix accessibility of unbindCompliance function

High : 
- [x] error message if burn value exceeds balance 
- [x] add doc on IModule about the action functions and why they may be empty
- [x]  add gaps at the end of storage contracts

Medium :
- [x] put a limit on looping functions to avoid hitting gas limit
- [x] Check Effects Interaction Violations
- [x] Validations on function calls 

Low :
- [x] Fix solidity version
- [x] remove useless variable in deploy function  
- [x] Fix functions visibility
- [x] Remove unused imports
- [x] Reorder functions in contracts following best practices
- [x] standard lint ==> double quotes for literals 
- [x] Add Missing Zero Address/String Input Check
- [x] NatSpec typo corrections
- [x] Error messages typo corrections
- [x] Unindexed events fix

Additional changes : 
- [x] Change implementationAuthority address for a token
- [x] modif of isVerified implem (#129) 
- [x] test coverage registries (#130)  
- [x] modifictions on IA (#131)  
- [x] lint all contracts (#132) 
- [x] achieve max coverage (#133)
- [x] use the ERC-1822 storage slot on proxies to avoid storage collision


contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -114,6 +114,7 @@ contract ModularCompliance is IModularCompliance, OwnableUpgradeable, MCStorage
      */
     function addModule(address _module) external override onlyOwner {
         require(!moduleBound[_module], 'module already bound');
+        require(modules.length <= 25, 'cannot add more than 25 modules');
```
Nakasar, line 117:
```suggestion
        require(modules.length <= 24, 'cannot add more than 25 modules');
```

Otherwise it means we cas add 26 modules.

Nakasar:
```suggestion
        require(modules.length < 25, 'cannot add more than 25 modules');
```

Ot strict comparison.

contracts/compliance/modular/modules/CountryRestrictModule.sol
```diff
@@ -124,9 +124,11 @@ contract CountryRestrictModule is AbstractModule {
      *  @param _countries Countries to be restricted, should be expressed by following numeric ISO 3166-1 standard
      *  Can be called only for a compliance contract that is bound to the CountryRestrict Module
      *  Only the owner of the Compliance smart contract can call this function
-     *  emits an `AddedRestrictedCountry` event
+     *  cannot restrict more than 195 countries in 1 batch
+     *  emits _countries.length `AddedRestrictedCountry` events
      */
     function batchRestrictCountries(uint16[] calldata _countries) external onlyComplianceCall {
+        require(_countries.length < 195, 'only 195 countries in the world');
```
Nakasar, line 131:
```suggestion
        require(_countries.length < 195, 'maximum 195 can be restricted, use unrestrict if needed');
```

Maybe not affirm how many countries there are in the world!

contracts/compliance/modular/modules/CountryRestrictModule.sol
```diff
@@ -138,10 +140,12 @@ contract CountryRestrictModule is AbstractModule {
      *  Identities from those countries will again be authorised to manipulate Tokens linked to this Compliance.
      *  @param _countries Countries to be unrestricted, should be expressed by following numeric ISO 3166-1 standard
      *  Can be called only for a compliance contract that is bound to the CountryRestrict Module
+     *  cannot unrestrict more than 195 countries in 1 batch
      *  Only the owner of the Compliance smart contract can call this function
-     *  emits an `RemovedRestrictedCountry` event
+     *  emits _countries.length `RemovedRestrictedCountry` events
      */
     function batchUnrestrictCountries(uint16[] calldata _countries) external onlyComplianceCall {
+        require(_countries.length < 195, 'only 195 countries in the world');
```
Nakasar, line 148:
```suggestion
        require(_countries.length < 195, 'maximum 195 can be unrestricted in one batch');
```

contracts/factory/TREXFactory.sol
```diff
@@ -138,8 +138,21 @@ contract TREXFactory is ITREXFactory, Ownable {
...
+        require((_claimDetails.issuers).length == (_claimDetails.issuerClaims).length
+        , 'claim pattern not valid');
+        require((_claimDetails.issuers).length <= 5
+        , 'max 5 claim issuers at deployment');
+        require((_claimDetails.claimTopics).length <= 5
+        , 'max 5 claim topics at deployment');
+        require((_tokenDetails.irAgents).length <= 5 && (_tokenDetails.tokenAgents).length <= 5
+        , 'max 5 agents at deployment');
+        require((_tokenDetails.complianceModules).length <= 30
+        , 'max 30 module actions at deployment');
```
Nakasar, line 152:
```suggestion
        require((_tokenDetails.complianceModules).length <= 25
        , 'max 25 module actions at deployment');
```

We limit 25 modules at compliance level I think.

contracts/registry/implementation/ClaimTopicsRegistry.sol
```diff
@@ -76,6 +76,7 @@ contract ClaimTopicsRegistry is IClaimTopicsRegistry, OwnableUpgradeable, CTRSto
      */
     function addClaimTopic(uint256 _claimTopic) external override onlyOwner {
         uint256 length = claimTopics.length;
+        require(length <= 15, 'cannot require more than 15 topics');
```
Nakasar, line 79:
```suggestion
        require(length < 15, 'cannot require more than 15 topics');
```

contracts/registry/implementation/IdentityRegistry.sol
```diff
@@ -205,6 +205,7 @@ contract IdentityRegistry is IIdentityRegistry, AgentRoleUpgradeable, IRStorage
             if (claimIds.length == 0) {
                 return false;
             }
+            require(claimIds.length <= 10, 'too much claims of same topic');
```
Nakasar, line 208:
```suggestion
            require(claimIds.length < 10, 'too much claims of same topic');
```

contracts/registry/implementation/IdentityRegistryStorage.sol
```diff
@@ -143,6 +143,7 @@ contract IdentityRegistryStorage is IIdentityRegistryStorage, AgentRoleUpgradeab
      *  @dev See {IIdentityRegistryStorage-bindIdentityRegistry}.
      */
     function bindIdentityRegistry(address _identityRegistry) external override {
+        require(identityRegistries.length <= 300, 'cannot bind more than 300 IR to 1 IRS');
```
Nakasar, line 146:
```suggestion
        require(identityRegistries.length < 300, 'cannot bind more than 300 IR to 1 IRS');
```

contracts/registry/implementation/TrustedIssuersRegistry.sol
```diff
@@ -80,6 +80,8 @@ contract TrustedIssuersRegistry is ITrustedIssuersRegistry, OwnableUpgradeable,
     function addTrustedIssuer(IClaimIssuer _trustedIssuer, uint256[] calldata _claimTopics) external override onlyOwner {
         require(trustedIssuerClaimTopics[address(_trustedIssuer)].length == 0, 'trusted Issuer already exists');
         require(_claimTopics.length > 0, 'trusted claim topics cannot be empty');
+        require(_claimTopics.length <= 15, 'cannot have more than 15 claim topics');
+        require(trustedIssuers.length <= 50, 'cannot have more than 50 trusted issuers');
```
Nakasar, line 84:
```suggestion
        require(trustedIssuers.length < 50, 'cannot have more than 50 trusted issuers');
```

contracts/factory/TREXFactory.sol
Joachim-Lebrun: it is not just about the max amount of modules here, but also about calls to module functions to set them up

contracts/compliance/modular/ModularCompliance.sol
Joachim-Lebrun: indeed

contracts/registry/implementation/IdentityRegistry.sol
```diff
@@ -205,7 +205,7 @@ contract IdentityRegistry is IIdentityRegistry, AgentRoleUpgradeable, IRStorage
             if (claimIds.length == 0) {
                 return false;
             }
-            require(claimIds.length <= 10, 'too much claims of same topic');
+            require(claimIds.length < 10, 'too much claims of same topic');
```
Nakasar, line 208: This one is important to note:
- A user with too much claims of the same topic will not be able to receive TREX tokens.
- The transaction would be reverted, and the user would have to remove some of their claims.

It's possible that many claims of the same topics exists on an identity. Let's say we have a claim of topic "YOU ARE OVER 18", then many issuers could emit such a claim.

Maybe we should think of another way to query claims, as follow:
- At this point in the contract, we know the topic (for `claimTopic` in `requiredClaimTopics`).
- We can query the list of claim topics associated with a trusted issuer (but I don't think we can query the list of trusted issuers for a given claim topic, we could optimize there).
- For each claim topic, we compute the claim ID with the trusted claim issuers.
- We can query on the identity contract each one of these computed claim ID.

Instead of querying an array of claims (that may be issued by untrusted claim issuers), we would only query claim of topic emitted by trusted issuers for this specific topic.

I can implement something like this to see how it goes.

Joachim-Lebrun: indeed, knowing that the claim ID is `bytes32 claimId = keccak256(abi.encode(_issuer, _topic))` we could check it this way instead of searching in an array of claims, it would cost less gas probably

Joachim-Lebrun: After checking closer, it is not better, or we should modify the TIR to have a function to get the list of trusted issuers for a given claim topic, otherwise you will need to call `hasClaimTopic` on each trusted issuer for each claim topic required, to calculate the claimID and look if the OID owner has that claimID and if it is valid and not revoked. 
Probably more costly than the actual implementation. 
If we want to go this way we need to add a variable `mapping(uint256 => address[]) trustedIssuers` and a function `getTrustedIssuers(uint256 _claimTopic)` ... It is not complicated to do, we can do it maybe... not sure if it would really save gas or not

Nakasar: In #129 I will implement the first optimization (computing the claim ID to query only claims from trusted issuers).
I might also implement a `getTrustedIssuersByClaimTopic(uint256 claimTopic)` method in the Trusted Issuers Registry.

PR #129, opened 2022-12-19
✨(identityRegistry) Compute array of claimIds from trusted issuers

Nakasar:
- This prevents from querying claims that would not be allowed because they were emitted from an untrusted issuer.
- This also prevents an Identity Owner to be unable to be verified because they might have too many claims of a given topic.

> **Note**
> Further optimization could be performed if the Trusted Issuer Registry had a method `getTrustedIssuersByClaimTopic(uint256 claimTopic)` (I might implement this on this PR).

- [x] modify `isVerified` function to check only claimIDs from trusted issuers
- [x] modify TrustedIssuersRegistry to store a mapping between claim topic and trusted issuers
- [x] add test cases for the modifications 
- [x] modify NatSpec for `isVerified`
- [x] add NatSpec for modifications on TIR 
- [x] update in changeLog 
- [x] (eventually) compare gas cost of new `isVerified` implem to old implem

Gas cost comparison for a transfer with and without this optimization (minor):

(comparing https://github.com/TokenySolutions/T-REX/actions/runs/3727061919/jobs/6320901174 and https://github.com/TokenySolutions/T-REX/actions/runs/3732257985/jobs/6331470948)

| Operation | With this PR | Without |
| *-------* | ------------ | ------- |
| Transfer  | 155742  | 191913 |
| Batch with 20 transfers | 2196700  | 2920120 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,20 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [4.0.0]
+
+### Breaking changes
+
+### Update
+- Trusted Issuers Registry Storage now maintains a mapping of truster issuers addresses allowed for a given claim topic.
+  - This means adding/removing/updating a trusted issuer now cost more gas (especially removing and updating when removing topics).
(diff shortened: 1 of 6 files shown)


Joachim-Lebrun (approved):
looks good to me :) we went down approx 40k gas units per transfer with the change, and it's only with 1 required claim!! very nice improvement

PR #130, opened 2022-12-19
✅(registries) Add test to cover missing statements

Nakasar:
Add tests to improve coverage.

- [x] Add test for `IdentityRegistry.isVerified` (statements that were not already covered by transfer tests).
- [x] Add tests for `IdentityRegistry.setIdentityRegistryStorage`.

We do have an issue with the try/catch that seems to not properly catch the error when the Trusted Issuer in the claim is not a Trusted Issuer contract (the `isVerified` returns a revert).

### DIFF
--- a/contracts/registry/implementation/IdentityRegistry.sol
+++ b/contracts/registry/implementation/IdentityRegistry.sol
@@ -215,8 +215,7 @@ contract IdentityRegistry is IIdentityRegistry, AgentRoleUpgradeable, IRStorage 
                         if (!_validity && j == (claimIds.length - 1)) {
                             return false;
                         }
-                    }
-                    catch {
+                    } catch {
                         if (j == (claimIds.length - 1)) {
                             return false;
                         }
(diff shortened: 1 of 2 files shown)


Joachim-Lebrun:
is the "try/catch" still necessary as we only check the validity of claimIDs corresponding to the required claim topic and the trusted issuer? The only case where it would be relevant in that case is if the token issuer adds an invalid trusted issuer, i.e. a trusted issuer address that does not correspond to a claim issuer contract and does not implement claim issuers methods

Nakasar:
Well, found the issue:

```
try IClaimIssuer(issuer).isClaimValid(identity(_userAddress), requiredClaimTopics[claimTopic], sig,
                        data) returns(bool _validity) {

                        if (
                            _validity
                        ) {
                            j = claimIds.length;
                        }
                        if (!_validity && j == (claimIds.length - 1)) {
                            return false;
                        }
                    }
                    catch {
                        if (j == (claimIds.length - 1)) {
                            return false;
                        }
                    }
```

Should have been
```solidity
try IClaimIssuer(issuer).isClaimValid(identity(_userAddress), requiredClaimTopics[claimTopic], sig,
                        data) returns(bool _validity) {

                        if (
                            _validity
                        ) {
                            j = claimIds.length;
                        }
                        if (!_validity && j == (claimIds.length - 1)) {
                            return false;
                        }
                    } catch {
                        if (j == (claimIds.length - 1)) {
                            return false;
                        }
                    }
```

That's all :D

PR #149, opened 2023-05-12
Features/fix config

Aboudjem:
Due to branching limitations, I combined all changes into one PR instead of separate ones for 'fix-config' and 'optimize-dvd'. All changes were merged to T-REX/develop. To ensure compliance with project standards, I ran the linter fix to pass the GitHub actions. This PR now contains all necessary changes for your review.

- [x] Fixing TypeScript linter 🔧
- [x] Prettier config update 💅
- [x] Gas reporter addition ⛽
- [x] Script fixes running on JS 🐛
- [x] Package additions for TypeScript, ESLint, and Prettier 📦
- [x] Package-lock update 🔐
- [x] TypeScript config fix 🔧
- [x] Fix linter warnings 🚨


.solhint.json
```diff
@@ -19,15 +19,16 @@
...
     "max-states-count": ["error", 15],
-    "no-empty-blocks": "error",
     "no-unused-vars": "error",
     "payable-fallback": "error",
     "constructor-syntax": "error",
     "not-rely-on-time": "off",
-    "reason-string": "off"
+    "reason-string": "off",
+    "no-global-import": "off",
+    "no-empty-blocks": "off"
```
Nakasar, line 31:
```suggestion
    "no-empty-blocks": "error"
```

This should remain as an error, ignore should be put where necessary (I think you added the rule because prettier changed how the functions were written, so the current ignores didn't work anymore, the `// solhint-disable-next-line no-empty-blocks` comment only needs to be moved a bit to be effective again, I'll open a PR to your branch to fix this!).

contracts/compliance/modular/modules/CountryRestrictModule.sol
```diff
@@ -153,21 +181,31 @@ contract CountryRestrictModule is AbstractModule {
      *  no transfer action required in this module
      */
     // solhint-disable-next-line no-empty-blocks
```
Nakasar, line 183: For instance, having the function written accross multiple line lines makes the `// solhint-disable-next-line no-empty-blocks` ineffective. It should be put a bit below.

Nakasar (changes requested):
Hey!
I'm not convice by the prettier/eslint configuration (usage of recommended node instead of typescript recommended). Also, some rules should not be ignore (I think you removed the rules because activating prettier broke a few ignore on the files).

I'm opening a PR to this branch with suggested corrections!

PR #167, opened 2023-09-14
✨(modules) Add name to modules

Nakasar:
- Add `function name() external pure returns (string memory);` to IModule. Compliance modules now require a `string constant public name = "ConditionalTransferModule";` constant variable to be declared.


Joachim-Lebrun:
should add for Max Balance Module that was just merged to the develop branch as well

PR #168, opened 2023-09-20
(TFM) Add transfer fees module

aliarbak:
Notes:
- The minimum fee rate is 0.01% but must be given in uint format. So, when defining fee rates:
  - 1 = 0.01%
  - 10 = 0.1%
  - 100 = 1%
  - 1000 = 10%
  - 10000 = 100%
  - It can also be 0, which means no fee will be charged
- `canComplianceBind` function is for the plug-and-play support discussed before. It checks whether the module is a token agent (and is required for the module to function). It should be called from the `addComplianceModule`
  - This is a draft function, I can update it after we decide how to proceed with the plug-and-play support.

### DIFF
--- a/contracts/compliance/modular/modules/TransferFeesModule.sol
+++ b/contracts/compliance/modular/modules/TransferFeesModule.sol
@@ -1,0 +1,210 @@
+// SPDX-License-Identifier: GPL-3.0
+//
+//                                             :+#####%%%%%%%%%%%%%%+
+//                                         .-*@@@%+.:+%@@@@@%%#***%@@%=
+//                                     :=*%@@@#=.      :#@@%       *@@@%=
+//                       .-+*%@%*-.:+%@@@@@@+.     -*+:  .=#.       :%@@@%-
+//                   :=*@@@@%%@@@@@@@@@%@@@-   .=#@@@%@%=             =@@@@#.
+//             -=+#%@@%#*=:.  :%@@@@%.   -*@@#*@@@@@@@#=:-              *@@@@+
+//            =@@%=:.     :=:   *@@@@@%#-   =%*%@@@@#+-.        =+       :%@@@%-
+//           -@@%.     .+@@@     =+=-.         @@#-           +@@@%-       =@@@@%:
+//          :@@@.    .+@@#%:                   :    .=*=-::.-%@@@+*@@=       +@@@@#.
+//          %@@:    +@%%*                         =%@@@@@@@@@@@#.  .*@%-       +@@@@*.
+//         #@@=                                .+@@@@%:=*@@@@@-      :%@%:      .*@@@@+
+//        *@@*                                +@@@#-@@%-:%@@*          +@@#.      :%@@@@-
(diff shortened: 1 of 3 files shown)


contracts/compliance/modular/modules/TransferFeesModule.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+    *  @dev Sets the fee rate and collector of the given compliance
+    *  @param _compliance is the compliance contract address
+    *  @param _rate is the rate of the fee (0.01% = 1, 1% = 100, 100% = 10000)
+    *  @param _collector is the collector wallet address
+    *  Only the owner of the token smart contract can call this function
+    *  Collector wallet address must be verified
+    */
+    function setFee(address _compliance, uint256 _rate, address _collector) external {
```
Joachim-Lebrun, line 105: all owner scoped functions should be called through the ModularCompliance function `callModuleFunction` in order to standardize the way we setup modules. some exceptions can exist, if for some reason it is not possible to achieve the desired behaviour this way, but when possible it is better to do it. Please update the function to be called by the modular compliance through `callModuleFunction`

contracts/compliance/modular/modules/TransferFeesModule.sol
```diff
@@ -0,0 +1,210 @@
...
+        address tokenAddress = IModularCompliance(_compliance).getTokenBound();
+        require(msg.sender == AgentRole(tokenAddress).owner(), "only token owner can call");
+
+        if (_rate > 10000) {
+            revert FeeRateIsOutOfRange(_compliance, _rate);
+        }
+
+        if (_collector == address(0)) {
```
Joachim-Lebrun, line 113: if the address is zero address `isVerified` that is called right after will return `false` therefore we can remove this check here

contracts/compliance/modular/modules/TransferFeesModule.sol
```diff
@@ -0,0 +1,210 @@
...
+        _fees[_compliance].collector = _collector;
+        emit FeeUpdated(_compliance, _rate, _collector);
+    }
+
+    /**
+    *  @dev See {IModule-moduleTransferAction}.
+    */
+    function moduleTransferAction(address _from, address _to, uint256 _value) external override onlyComplianceCall {
+        if (_from == address(0) || _to == address(0)) {
+            return;
```
Joachim-Lebrun, line 132: i believe this part is unreachable as the token contract will never call the `transferred` function with zero address as `_from` or `_to`, we can remove this check imo

Joachim-Lebrun (changes requested):
- add the `isPlugAndPlay` and `canComplianceBind` after #169 is merged to develop
- remove useless checks to save gas
- modify `setFee` function to be called by the Modular Compliance contract

Joachim-Lebrun:
just need to still add the `isPlugAndPlay` and `canComplianceBind` functions, which are both always returning `true`

aliarbak:
Hey @Joachim-Lebrun, I need to rebase from develop, can you allow me to force-push to the transfer-fees-module branch?

Joachim-Lebrun (approved):

PR #169, opened 2023-09-21
Add plug & play functions to compliance modules

aliarbak:


contracts/compliance/modular/modules/MaxBalanceModule.sol
```diff
@@ -105,6 +110,21 @@ contract MaxBalanceModule is AbstractModule {
...
+     *  @param _id the ONCHAINID address of the token holder
+     *  @param _balance the current balance of the token holder
+     *  Only the owner of the Compliance smart contract can call this function
+     *  emits a `IDBalancePreSet` event
+     */
+    function preSetModuleState(address _compliance, address _id, uint256 _balance) external {
+        require(OwnableUpgradeable(_compliance).owner() == msg.sender, "only compliance owner call");
+        require(!IModularCompliance(_compliance).isModuleBound(address(this)), "cannot do on bound compliance");
+        _preSetModuleState(_compliance, _id, _balance);
+        _compliancePresetStatus[_compliance] = true;
```
Joachim-Lebrun, line 125: i would set the preset status to true only when all ID balances have been set, let's add an additional function to set the preset status to true.

contracts/compliance/modular/modules/MaxBalanceModule.sol
```diff
@@ -117,9 +137,18 @@ contract MaxBalanceModule is AbstractModule {
         address _compliance,
```
Joachim-Lebrun, line 137: depending on the amount of Identities on the token we try to preset it could be required to call multiple batches to avoid gas overflow (max gas on EVM blockchains is somewhere between 7 and 25M units, so you cannot put an infinite amount of elements in the array, would be interesting to calculate the average amount of gas and set a limit of array length on the function call to avoid the out of gas error at the start of the function instead of when the function already consumed the max gas amount)

Joachim-Lebrun (changes requested):

contracts/compliance/modular/modules/MaxBalanceModule.sol
```diff
@@ -105,6 +110,21 @@ contract MaxBalanceModule is AbstractModule {
...
+    /**
+     *  @dev pre-set the balance of a token holder per ONCHAINID
+     *  @param _compliance the address of the compliance contract to preset
+     *  @param _id the ONCHAINID address of the token holder
+     *  @param _balance the current balance of the token holder
+     *  Only the owner of the Compliance smart contract can call this function
+     *  emits a `IDBalancePreSet` event
+     */
+    function preSetModuleState(address _compliance, address _id, uint256 _balance) external {
+        require(OwnableUpgradeable(_compliance).owner() == msg.sender, "only compliance owner call");
```
Joachim-Lebrun, line 122: better to create an error than a require

contracts/compliance/modular/modules/MaxBalanceModule.sol
```diff
@@ -105,6 +110,21 @@ contract MaxBalanceModule is AbstractModule {
...
+        require(!IModularCompliance(_compliance).isModuleBound(address(this)), "cannot do on bound compliance");
```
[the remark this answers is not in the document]
aliarbak: done

Joachim-Lebrun (approved):

PR #171, opened 2023-10-05
(DvA) Add DvA Transfer Manager

aliarbak:


contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,521 @@
...
+        address tokenAddress;
+        address sender;
+        address recipient;
+        uint256 amount;
+        TransferStatus status;
+        Approver[] approvers;
+    }
+
+    struct Approver {
+        address wallet; // address(0) = any token agent
```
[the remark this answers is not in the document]
aliarbak: sure, I'm on it 👍

Nakasar:
We store the list of approvers in the `_transfers` mapping. This means if a user initiates a transfer, then the approval criteria are modified, then the transfer would still use the criteria at the date it was initiated and not the updated ones. Is this expected (I couldn't see any mention of this in the specs so maybe not)?

aliarbak:
> We store the list of approvers in the `_transfers` mapping. This means if a user initiates a transfer, then the approval criteria are modified, then the transfer would still use the criteria at the date it was initiated and not the updated ones. Is this expected (I couldn't see any mention of this in the specs so maybe not)?

Not sure, I'll ask 👍

aliarbak:
> We store the list of approvers in the `_transfers` mapping. This means if a user initiates a transfer, then the approval criteria are modified, then the transfer would still use the criteria at the date it was initiated and not the updated ones. Is this expected (I couldn't see any mention of this in the specs so maybe not)?

The final business decision here: If approval criteria change after a transfer is initiated, we need to reset the transfer's approval state (and invalidate all previous approvers).

And when an approver approves a transfer after the criteria change, it needs to reset the approval state and return. S/he needs to call the approve function one more time to be able to approve. This is the expected behavior.  
(I updated the contract btw)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
   constant variable to be declared.
 - Add `function isPlugAndPlay() external pure returns (bool)` to `IModule`. Compliance modules now require this function to be declared. It indicates whether the compliance can be bound without any presetting.
 - Add `function canComplianceBind(address _compliance) external view returns (bool)` to `IModule`. Compliance modules now require this function to be declared. If it returns false, it means some presets must be made before compliance can bind to the module.
+- DVATransferManager contract
```
Joachim-Lebrun, line 19: would be good to describe the contract a bit here

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+contract DVATransferManager {
+    enum TransferStatus {
+        PENDING,
+        COMPLETED,
+        CANCELLED,
+        REJECTED
+    }
+
+    struct ApprovalCriteria {
+        address tokenAddress;
```
Joachim-Lebrun, line 77:
```suggestion
```

As the ApprovalCriteria is given by a `mapping(address => ApprovalCriteria)` with `address` being the token address, it feels redundant to have it in the struct

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+     *  @param tokenAddress is the token address.
+     *  @param includeRecipientApprover determines whether the recipient is included in the approver list
+     *  @param includeAgentApprover determines whether the agent is included in the approver list
+     *  @param sequentialApproval determines whether approvals must be sequential
+     *  @param additionalApprovers are the addresses of additional approvers to be added to the approver list
+     *  Only an agent of a token can call this function
+     *  DVATransferManager must be an agent of the given token
+     *  emits an `ApprovalCriteriaModified` event
+     */
+    function modifyApprovalCriteria(
```
Joachim-Lebrun, line 254: not sure about the naming here. `modify` implies that the `ApprovalCriteria` already exists, while in fact this function is also used to set initial `ApprovalCriteria` 
In this case i guess the terminology `setApprovalCriteria` would be more relevant

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    /**
+    *  @dev modify the approval criteria of a token
+     *  @param tokenAddress is the token address.
+     *  @param includeRecipientApprover determines whether the recipient is included in the approver list
+     *  @param includeAgentApprover determines whether the agent is included in the approver list
+     *  @param sequentialApproval determines whether approvals must be sequential
+     *  @param additionalApprovers are the addresses of additional approvers to be added to the approver list
+     *  Only an agent of a token can call this function
```
Joachim-Lebrun, line 250: Agents on T-REX are in charge of operational tasks such as minting, burning, freezing, recovering lost tokens, etc. Here i believe the task is more related to the general transfer rules management for the token, which is under the owner scope in T-REX, in the same way that it is the token owner that specifies the compliance rules, or the eligibility rules, i believe it should be the Token owner specifying the approval criterias in DvA

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    /**
+     *  @dev initiates a new transfer
+     *  @param tokenAddress is the address of the token
+     *  @param recipient is the address of the recipient
+     *  @param amount is the transfer amount
+     *  Approval criteria must be preset for the given token address
+     *  Receiver must be verified for the given token address
+     *  emits a `TransferInitiated` event
```
Joachim-Lebrun, line 303: could add that the sender should give allowance first to the DvA smart contract address on the token contract, and that this allowance needs to be at least `amount`

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+     */
+    function initiateTransfer(address tokenAddress, address recipient, uint256 amount) external {
+        ApprovalCriteria memory approvalCriteria = _approvalCriteria[tokenAddress];
+        if (approvalCriteria.tokenAddress == address(0)) {
+            revert TokenIsNotRegistered(tokenAddress);
+        }
+
+        IToken token = IToken(tokenAddress);
+        if (!token.identityRegistry().isVerified(recipient)) {
+            revert RecipientIsNotVerified(tokenAddress, recipient);
```
Joachim-Lebrun, line 313: maybe would make sense to also test `canTransfer` with address `_from` being the DvA address and `_to` the recipient for `amount`, to be sure that after approval, the transfer will be allowed to happen. Or maybe this is a bit overkill...

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+        }
+
+        bool transferSent = token.transferFrom(msg.sender, address(this), amount);
+        if (!transferSent) {
+            revert TokenTransferFailed(tokenAddress, msg.sender, address(this), amount);
```
Joachim-Lebrun, line 318: I don't think this code is reachable. When transfer fails it reverts, therefore i don't see how we could get the case where `transferFrom` returns `false` and reach this code. I would therefore remove this part and the custom error that goes with it

contracts/DVA/DVATransferManager.sol
```diff
@@ -0,0 +1,651 @@
...
+
+        for (uint256 i = 0; i < approvalCriteria.additionalApprovers.length; i++) {
+            transfer.approvers.push(Approver(approvalCriteria.additionalApprovers[i], false, false));
+        }
+    }
+
+    function _transferTokensTo(Transfer memory transfer, address to) internal {
+        bool transferSent = IToken(transfer.tokenAddress).transfer(to, transfer.amount);
+        if (!transferSent) {
+            revert TokenTransferFailed(transfer.tokenAddress, address(this), transfer.sender, transfer.amount);
```
Joachim-Lebrun, line 626: same remark as for `transferFrom` above, this code is not reachable, `transfer` either returns `true` or reverts. therefore it is not possible to get access to this case in which the transfer returns `false` and reach this custom error

Joachim-Lebrun:
@aliarbak we should also have an implementation contract for DvA

contracts/DVA/DVATransferManager.sol
aliarbak: I also used it to check if the token is registered on the DvA or not
https://github.com/TokenySolutions/T-REX/pull/171/files#diff-162d3949e79668025b8b4e6bb7dadb1e3828224cd56852f69b2de7d4c4afd77aR134

We can't use other values in the ApprovalCriteria struct for this purpose. Or we could add a flag instead of keeping the tokenAddress, what do you think?

contracts/DVA/DVATransferManager.sol
aliarbak: We already have an e2e test for that: https://github.com/TokenySolutions/T-REX/pull/171/files#diff-5bfef1f607b3b2d5e5aa8f381f378a17e93375f7cce6bfd63dad06dc354c5b97R511

aliarbak:
all done @Joachim-Lebrun

PR #175, opened 2023-10-30
add gas reporter

Joachim-Lebrun:
gas reporter helps to see the gas use of each function in the T-REX suite of contracts and improve performance when necessary

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -15,6 +15,9 @@ const config: HardhatUserConfig = {
       },
     },
   },
+  gasReporter: {
+    enabled: true,
+  },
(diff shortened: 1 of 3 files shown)


Joachim-Lebrun:
we used eth-gas-reporter in the past, then removed it, but it seems it improved a lot in the last few years, it was quite useless before as it was just giving the gas cost of unit tests in the console, which was not useful as we almost always do multiple function calls in a single unit test, but now it outputs a table with the gas cost of each method call in the console when launching tests, therefore i think it is way more useful than before and should be re-added

Nakasar (approved):

PR #187, opened 2024-02-12
Modify IA storage slot on proxies

Joachim-Lebrun:
Proxies were using the storage slot from the ERC-1822 proxy pattern, but instead of using it to store direct implementation we used it to store the address of the Implementation Authority. 
The fact of using this storage slot created issues with the explorers such as polygonscan which believed we used ERC-1822 and displayed the ABI of the Implementation Authority instead of the ABI of the implementation contracts. 

This PR proposes a new storage slot `0x821f3e4d3d679f19eacc940c87acf846ea6eae24a63058ea750304437a62aafc` which is the result of `keccak256("ERC-3643.proxy.beacon")`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [4.1.3]
+
+### Update
+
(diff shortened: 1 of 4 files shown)


aliarbak (approved):
🚀

PR #189, opened 2024-02-13
Add pause and unpause functions to TREX Gateway

aliarbak:
- **TREXGateway Enhancements**:
  - New function `pause`: Pauses the TREX gateway contract. When paused, no one can deploy a new token.
  - New function `unpause`: Unpauses the TREX gateway contract.

### DIFF
--- a/contracts/factory/ITREXGateway.sol
+++ b/contracts/factory/ITREXGateway.sol
@@ -102,6 +102,12 @@ interface ITREXGateway {
     /// event emitted whenever a TREX token has been deployed by the TREX factory through the use of the Gateway
     event GatewaySuiteDeploymentProcessed(address indexed requester, address intendedOwner, uint256 feeApplied);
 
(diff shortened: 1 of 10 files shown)


contracts/factory/ITREXGateway.sol
```diff
@@ -102,6 +102,12 @@ interface ITREXGateway {
     /// event emitted whenever a TREX token has been deployed by the TREX factory through the use of the Gateway
     event GatewaySuiteDeploymentProcessed(address indexed requester, address intendedOwner, uint256 feeApplied);
 
+    /// event emitted whenever the gateway is paused by an admin
+    event GatewayPaused(address _userAddress);
```
aliarbak, line 106: I didn't use the names "Paused" and "Unpaused" because they are used by the token contract. Otherwise, sharing the same event names as token events would cause ambiguity and collision on the indexer side.

PR #202, opened 2024-05-26
♻(dependencies) upgrade ethers 6 and other dependencies

anonkey:
I upgraded all dependencies for my personal usage and thought it'll avoid you do the same thing on your own.

I restored the scripts/commit-msg.js because husky hooks was broken but maybe what was wished is to remove commitlint ?

And the override in package.json are here because of an empty package release published for these package, it will be removed as soon as the release is fixed

If you're okay i'm ready to do PR on [erc3643-packages](https://github.com/ERC-3643/erc3643-packages) too to avoid breaking it

### DIFF
--- a/.github/workflows/publish-prerelease.yml
+++ b/.github/workflows/publish-prerelease.yml
@@ -6,13 +6,13 @@ on:
 
 jobs:
   build:
-    if: 'github.event.release.prerelease'
+    if: "github.event.release.prerelease"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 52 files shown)


Joachim-Lebrun:
Thanks a lot for the work done here @anonkey ! We'll add this in the v4.2 of T-REX

Joachim-Lebrun:
i will solve the test error on branch 4.2.0 instead of doing it working on your fork of the repository, will be easier for me, that's why i merged even if a test was failing

anonkey:
Ok thank you

PR #204, opened 2024-06-14
Proxify DvA manager and implement token freeze logic

aliarbak:
- DvA Transfer Manager contract proxified
- The DvA manager contract freezes the tokens to be transferred instead of being a vault (so it must be a token agent) 
- Only the token owner (rather than agents) can call setApprovalCriteria, as it is part of the main token settings.

### DIFF
--- a/contracts/DVA/DVATransferManager.sol
+++ b/contracts/DVA/DVATransferManager.sol
@@ -436,10 +436,10 @@ contract DVATransferManager is IDVATransferManager, Initializable, OwnableUpgrad
         return transfer;
     }
 
+    // solhint-disable-next-line no-empty-blocks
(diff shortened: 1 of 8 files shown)


contracts/DVA/DVATransferManager.sol
```diff
@@ -77,7 +80,8 @@ contract DVATransferManager is IDVATransferManager {
     // nonce of the transaction allowing the creation of unique transferID
     uint256 private _txNonce;
 
-    constructor(){
+    function initialize() external initializer {
```
Joachim-Lebrun, line 83: can you add a test to see that you cannot call `initialize()` post deployment ?

PR #205, opened 2024-06-18
Trex 121 recovery function modify

mplelis:
https://tokeny.atlassian.net/browse/TREX-121


contracts/registry/implementation/IdentityRegistry.sol
```diff
@@ -268,8 +268,11 @@ contract IdentityRegistry is IIdentityRegistry, AgentRoleUpgradeable, IRStorage
         IIdentity _identity,
         uint16 _country
     ) public override onlyAgent {
-        _tokenIdentityStorage.addIdentityToStorage(_userAddress, _identity, _country);
-        emit IdentityRegistered(_userAddress, _identity);
+        IIdentity identity = _tokenIdentityStorage.storedIdentity(_userAddress);
```
Joachim-Lebrun, line 271: you shouldn't change the registerIdentity method for updating the recovery process

Joachim-Lebrun: if a wallet is already registered there is a function update to change the OID linked to it, add function shouldn't be used for that

test/token/token-recovery.test.ts
```diff
@@ -40,20 +40,6 @@ describe('Token - Recovery', () => {
         });
       });
 
-      describe('when new wallet is not authorized on the identity', () => {
```
Joachim-Lebrun, line 43: We need to test the case in which 2 tokens that are linked to the same IRS are doing the recovery as it is what the bug was.

Joachim-Lebrun: in old implementation it should have worked on the first token and then failed on the second because the Identity Registry storage was already modified by the first recovery.

Joachim-Lebrun (changes requested):

PR #210, opened 2024-06-28
TREX-132 use custom errors in require clauses

pgonday:
***** ABANDONNED PR *****
Replaced by https://github.com/TokenySolutions/T-REX/pull/211


contracts/DVD/DVDTransferManager.sol
```diff
@@ -155,6 +157,18 @@ contract DVDTransferManager is Ownable {
         address fee1Wallet,
         address fee2Wallet);
 
+    /// Errors
+
+    error FeeWalletCannotBeZeroAddress(uint256 walletIndex);
```
Joachim-Lebrun, line 162: should declare errors at file level (outside of contract brackets)

contracts/DVD/DVDTransferManager.sol
```diff
@@ -193,22 +207,22 @@ contract DVDTransferManager is Ownable {
...
-            _fee2 <= 10**_feeBase && _fee2 >= 0 &&
+            _fee1 <= 10**_feeBase && 
+            _fee2 <= 10**_feeBase && 
             _feeBase <= 5 &&
             _feeBase >= 2
-            , "invalid fee settings");
+            , InvalidFeeSettings());
         if (_fee1 > 0) {
-            require(_fee1Wallet != address(0), "fee wallet 1 cannot be zero address");
+            require(_fee1Wallet != address(0), FeeWalletCannotBeZeroAddress(1));
```
Joachim-Lebrun, line 222: why not use `ZeroAddress` error from your invalid arguments library?

contracts/DVD/DVDTransferManager.sol
```diff
@@ -193,22 +207,22 @@ contract DVDTransferManager is Ownable {
...
         }
         if (_fee2 > 0) {
-            require(_fee2Wallet != address(0), "fee wallet 2 cannot be zero address");
+            require(_fee2Wallet != address(0), FeeWalletCannotBeZeroAddress(2));
```
Joachim-Lebrun, line 225: same remark

contracts/DVD/DVDTransferManager.sol
```diff
@@ -250,12 +264,14 @@ contract DVDTransferManager is Ownable {
...
-        require (_counterpart != address(0), "counterpart cannot be null");
-        require(IERC20(_token2).totalSupply() != 0, "invalid address : address is not an ERC20");
+        uint256 balance = IERC20(_token1).balanceOf(msg.sender);
+        require(balance >= _token1Amount, CommonErrors.ERC20InsufficientBalance(msg.sender, balance, _token1Amount));
+
+        uint256 allowance = IERC20(_token1).allowance(msg.sender, address(this));
+        require(allowance >= _token1Amount, CommonErrors.ERC20InsufficientAllowance(msg.sender, allowance, _token1Amount));
+
+        require (_counterpart != address(0), InvalidArgumentErrors.ZeroAddress());
+        require(IERC20(_token2).totalSupply() != 0, InvalidArgumentErrors.AddressNotERC20(_token2));
```
Joachim-Lebrun, line 274: in facts i am not sure this error can throw. In case the address is not an ERC20 the call to `totalSupply()` will revert, not return 0 i think. In one way it is good as it reverts, as expected, but it will not revert with the error we want. Can solve this with a `try/catch` block i guess

contracts/DVD/DVDTransferManager.sol
```diff
@@ -314,20 +330,25 @@ contract DVDTransferManager is Ownable {
...
-            , "Not enough tokens in balance");
+            balance >= token2.amount
+            , CommonErrors.ERC20InsufficientBalance(token2.counterpart, balance, token2.amount));
+
+        uint256 allowance = token2Contract.allowance(token2.counterpart, address(this));
         require(
-            token2Contract.allowance(token2.counterpart, address(this)) >= token2.amount
-            , "not enough allowance to transfer");
+            allowance >= token2.amount
+            , CommonErrors.ERC20InsufficientAllowance(token2.counterpart, allowance, token2.amount));
```
Joachim-Lebrun, line 350: not sure this one is necessary, as the allowance is checked by the transferFrom a few lines under

contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -60,24 +60,41 @@
...
 import "./MCStorage.sol";
 import "./modules/IModule.sol";
+import "../../libraries/errors/InvalidArgumentErrors.sol";
 
 
 contract ModularCompliance is IModularCompliance, OwnableUpgradeable, MCStorage {
 
+    /// errors
+
+    error AddressNotATokenBoundToComplianceContract();
```
Joachim-Lebrun, line 77: declare the error at file level, outside of contract brackets

contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -60,24 +60,41 @@
...
+    error TokenNotBound();
+
     /// modifiers
 
     /**
      * @dev Throws if called by any address that is not a token bound to the compliance.
      */
     modifier onlyToken() {
-        require(msg.sender == _tokenBound, "error : this address is not a token bound to the compliance contract");
+        require(msg.sender == _tokenBound, AddressNotATokenBoundToComplianceContract());
```
Joachim-Lebrun, line 97: should use one of your compliance errors from lib

contracts/libraries/errors/CommonErrors.sol
```diff
@@ -0,0 +1,126 @@
...
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+pragma solidity 0.8.26;
+
+
+library CommonErrors {
+
+    error InitializationFailed();
```
Joachim-Lebrun, line 70: errors can be declared outside of library brackets, this way you don't need to call them as `CommonErrors.errorName` but just `errorName` when you import the file containing the errors

contracts/libraries/errors/ComplianceErrors.sol
```diff
@@ -0,0 +1,79 @@
...
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+pragma solidity 0.8.26;
+
+
+library ComplianceErrors {
```
Joachim-Lebrun, line 68: errors can be declared outside of library brackets, this way you don't need to call them as `CommonErrors.errorName` but just `errorName` when you import the file containing the errors

contracts/libraries/errors/InvalidArgumentErrors.sol
```diff
@@ -0,0 +1,85 @@
...
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+pragma solidity 0.8.26;
+
+
+library InvalidArgumentErrors {
```
Joachim-Lebrun, line 67: errors can be declared outside of library brackets, this way you don't need to call them as `CommonErrors.errorName` but just `errorName` when you import the file containing the errors

contracts/libraries/errors/RoleErrors.sol
```diff
@@ -0,0 +1,99 @@
...
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+pragma solidity 0.8.26;
+
+
+library RoleErrors {
```
Joachim-Lebrun, line 67: errors can be declared outside of library brackets, this way you don't need to call them as `CommonErrors.errorName` but just `errorName` when you import the file containing the errors

contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -111,12 +127,12 @@ contract ModularCompliance is IModularCompliance, OwnableUpgradeable, MCStorage
...
-        require(_module != address(0), "invalid argument - zero address");
-        require(!_moduleBound[_module], "module already bound");
-        require(_modules.length <= 24, "cannot add more than 25 modules");
+        require(_module != address(0), InvalidArgumentErrors.ZeroAddress());
+        require(!_moduleBound[_module], ModuleAlreadyBound());
+        require(_modules.length <= 24, CannotAddMoreThan25Modules());
         IModule module = IModule(_module);
         if (!module.isPlugAndPlay()) {
-            require(module.canComplianceBind(address(this)), "compliance is not suitable for binding to the module");
+            require(module.canComplianceBind(address(this)), ComplianceNotSuitableForBindingToModule());
```
Joachim-Lebrun, line 135: i would add the module address as a parameter of this error, this way you know which module cannot be bound even if you try to bind a batch of modules (which is what TREX Factory does at deployment for example)

contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -60,24 +60,41 @@
...
 
+    /// errors
+
+    error AddressNotATokenBoundToComplianceContract();
+
+    error CannotAddMoreThan25Modules();
+
+    error ComplianceNotSuitableForBindingToModule();
+
+    error ModuleAlreadyBound();
```
Joachim-Lebrun, line 83: some of these errors should be imported from the compliance errors library i guess, these are very common to modules and compliance contracts

contracts/compliance/modular/ModularCompliance.sol
```diff
@@ -198,24 +214,14 @@ contract ModularCompliance is IModularCompliance, OwnableUpgradeable, MCStorage
...
-            call(
-            gas(),
-            _module,
-            0,
-            freeMemoryPointer,
-            callData.length,
-            0,
-            0
-            ))
+            if iszero(call(gas(), _module, 0, freeMemoryPointer, callData.length, 0, 0))
```
Joachim-Lebrun, line 217: not sure it is more lisible like this, instead i would keep the formatting we had before and maybe add a comment on each line to explain what the assembly does, people don't understand assembly so it is good to document it

Joachim-Lebrun: could be 
```solidity
assembly {
    let freeMemoryPointer := mload(0x40) // Load the free memory pointer from memory location 0x40
    calldatacopy(freeMemoryPointer, callData.offset, callData.length) // Copy callData from calldata to the free memory location
    if iszero( // Check if the call returns zero (indicating failure)
        call( // Perform the external call
            gas(), // Provide all available gas
            _module, // Address of the target module
            0, // No ether is sent with the call
            freeMemoryPointer, // Input data starts at the free memory pointer
            callData.length, // Input data length
            0, // Output data location (not used)
            0 // Output data size (not used)
        )
    ) {
        returndatacopy(0, 0, returndatasize()) // Copy return data to memory starting at position 0
        revert(0, returndatasize()) // Revert the transaction with the return data
    }
}
```

contracts/compliance/modular/modules/ConditionalTransferModule.sol
```diff
@@ -100,6 +100,10 @@ contract ConditionalTransferModule is AbstractModuleUpgradeable {
      */
     event ApprovalRemoved(address _from, address _to, uint _amount, address _token);
 
+    /// Errors
+
+    error NotApproved();
```
Joachim-Lebrun, line 105: could add transfer details, in case the transaction is part of a batch it would help to debug to know which element of the batch is causing an issue.

contracts/compliance/modular/modules/MaxBalanceModule.sol
```diff
@@ -107,6 +107,8 @@ contract MaxBalanceModule is AbstractModuleUpgradeable {
 
     error TokenAlreadyBound(address _compliance);
 
+    error IdentityNotFound();
```
Joachim-Lebrun, line 110: i would add the wallet address for which identity was not found as a parameter here, would help to debug in case of batch transfers and know which sub-transfer failed

PR #215, opened 2024-07-30
TREX-129 default allowance from develop

pgonday:


contracts/token/Token.sol
```diff
@@ -385,6 +387,15 @@ contract Token is IToken, AgentRoleUpgradeable, TokenStorage {
         revert RecoveryNotPossible();
     }
 
+    /// @dev See {IToken-setAllowanceForAll}.
+    function setAllowanceForAll(bool _allow, address _caller, address[] calldata _targets) external override onlyAgent {
```
Joachim-Lebrun, line 391: scope of the function should be onlyOwner, not agent

contracts/token/Token.sol
```diff
@@ -385,6 +387,15 @@ contract Token is IToken, AgentRoleUpgradeable, TokenStorage {
...
+        uint256 targetsCount = _targets.length;
+        for (uint256 i = 0; i < targetsCount; i++) {
+            _defaultAllowances[_caller][_targets[i]] = _allow;
+            emit DefaultAllowance(_caller, _targets[i], _allow);
+        }
+    }
+
```
Joachim-Lebrun, line 398: we should provide a function for token holders to get out of the default allowance if they want to handle allowance at their level, keeping more control this way

contracts/token/Token.sol
```diff
@@ -385,6 +387,15 @@ contract Token is IToken, AgentRoleUpgradeable, TokenStorage {
         revert RecoveryNotPossible();
     }
```
Joachim-Lebrun, line 389: to keep compatibility with most DeFi platforms and other external tools we cannot have something so custom, we need to respect ERC20, therefore the function `allowance` should be overriden to return max(uint256) when defaultAllowance is set for that spender address and that the owner didn't opt out of the default allowance (we should provide a function for token holder to manage allowance normally and refuse the default allowance if they deem it preferable)

Joachim-Lebrun: in current implementation the `allowance` would return for example 0 while the transferFrom is going to pass, it is going to create a lot of bugs as DeFi platforms are implemented to ask you to increase allowance if `allowance` returns an amount lower than the amount to transfer

Joachim-Lebrun (approved):

Joachim-Lebrun (approved):

PR #218, opened 2024-08-05
TREX-14 Eligibility & Compliance checker contract

pgonday:


package.json
```diff
@@ -85,5 +85,8 @@
     "*.js": [
       "eslint"
     ]
+  },
+  "dependencies": {
+    "hardhat-tracer": "^3.0.3"
```
Nakasar, line 90: Should be in `devDependencies`

```bash
npm uninstall hardhat-tracer && npm install --save-dev hardhat-tracer
```

contracts/utilities/IUtilityChecker.sol
```diff
@@ -0,0 +1,123 @@
...
+    ///     - check if token is paused,
+    ///     - check freeze conditions,
+    ///     - check compliance.
+    /// @param _compliance Address of the Modular Compliance contract.
+    /// @param _token The address of the token contract.
+    /// @param _from The address of the sender.
+    /// @param _to The address of the recipient.
+    /// @param _amount The amount of tokens to be transferred.
+    /// @return bool Returns true if the transfer would be successful, false otherwise.
+    function testTransfer(address _compliance, address _token, address _from, address _to, uint256 _amount)
```
Joachim-Lebrun, line 101: the interface can be simplified by removing the compliance address, this address can be found easily if you have the token address as there is a getter on the token interface to get the compliance address.

Joachim-Lebrun:
@pgonday can you solve the conflicts on the PR please ?

contracts/utilities/UtilityChecker.sol
```diff
@@ -0,0 +1,179 @@
...
+contract UtilityChecker is IUtilityChecker, OwnableUpgradeable, UUPSUpgradeable {
+
+    function initialize() external initializer {
+        __Ownable_init();
+    }
+
+    /// @inheritdoc IUtilityChecker
+    /// @dev This function is not gas optimized and should be called only OFF chain.
+    function testTransfer(address _token, address _from, address _to, uint256 _amount) 
+        external view override returns (bool) {
```
Joachim-Lebrun, line 86: we could make this function return 3 boolean values 
```suggestion
        external view override returns (bool _freezeStatus, bool _eligibilityStatus, bool _complianceStatus) {
```
This way the user knows immediately where to check to get the explanation of why the transfer is going to fail

contracts/utilities/UtilityChecker.sol
```diff
@@ -0,0 +1,179 @@
...
+        IToken token = IToken(_token);
+
+        if (token.paused()) return false;
+
+        (bool frozen, ) = testFreeze(_token, _from, _to, _amount);
+        if (frozen) return false;
+
+        ComplianceCheckDetails [] memory details = testTransferDetails(_token, _from, _to, _amount);
```
Joachim-Lebrun, line 94: you check compliance, pause and freeze but not the eligibility in this function, which means that right now a transfer that would block at the eligibility level will return true, while it should return false. You need to check the eligibility on the receiver address (just need to call `isVerified` on the IR of the token

Joachim-Lebrun:
```suggestion
        IIdentityRegistry ir = token.identityRegistry();
        if(!ir.isVerified(_to)) return false; 
        ComplianceCheckDetails [] memory details = testTransferDetails(_token, _from, _to, _amount);
```

Joachim-Lebrun (changes requested):

pgonday:
Relaced by https://github.com/TokenySolutions/T-REX/pull/229

PR #225, opened 2024-09-10
Add preset completed event to max balance module

aliarbak:
- Added **PresetCompleted** event to MaxBalanceModule, which is triggered when the presetCompleted function is called.


pgonday (commented):
lgtm

Joachim-Lebrun (approved):

PR #226, opened 2024-09-17
Ownership2steps

pgonday:
Change OwnableUpgradeable to Ownable2StepUpgradeable
https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.9.6/contracts/access/Ownable2StepUpgradeable.sol


Joachim-Lebrun:
please update the changelog file to add the modifications

PR #233, opened 2024-12-06
BT-129 - IRS ERC3643 implementation

mplelis:


contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
```diff
@@ -0,0 +1,173 @@
...
+     */
+    function removeIdentityFromStorage() external view override onlyAgent {
+        revert ContractIsReadOnly();
+    }
+
+    /**
+     *  @dev See {IIdentityRegistryStorage-bindIdentityRegistry}.
+     */
+    function bindIdentityRegistry() external view override {
+        revert ContractIsReadOnly();
```
Joachim-Lebrun, line 133: just do nothing, don't revert, otherwise the factory won't be able to use this contract, as the factory will try to bind

contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
```diff
@@ -0,0 +1,173 @@
...
+     */
+    function storedIdentity(address _userAddress) external view override returns (IIdentity) {
+        return IIdentity(_iidFactory.getIdentity(_userAddress));
+    }
+
+    /**
+     *  @dev See {IIdentityRegistryStorage-storedInvestorCountry}.
+     */
+    function storedInvestorCountry(address _userAddress) external view override returns (uint16) {
+        return _identities[_userAddress].investorCountry;
```
Joachim-Lebrun, line 161: just return 42 or something, by default, in case the `_identities[_userAddress].investorCountry` value is 0 
Should have a way for OID owners to publish their country code there in a decentralized way, so that a wallet that is linked to an OID on the IdFactory can update their own country code on this contract

contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
```diff
@@ -0,0 +1,173 @@
...
+
+/// @dev Thrown when trying to call write functions.
+error ContractIsReadOnly();
+
+/// @dev Thrown when maximum numbe of identity registry by identity registry storage is reached.
+/// @param _max miximum number of IR by IRS.
+error MaxIRByIRSReached(uint256 _max);
+
+
+contract IdentityRegistryStorage is IERC3643IdentityRegistryStorage, AgentRoleUpgradeable, IRSStorage, IERC165 {
```
Joachim-Lebrun, line 91: call this contract another way, we already have an identity registry storage

contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
```diff
@@ -0,0 +1,173 @@
...
+    function init(_identityFactoryAddress) external initializer {
+        require(_identityFactoryAddress != address(0), ZeroAddress());
+        _iidFactory = IIdFactory(_identityFactoryAddress);
+        __Ownable_init();
+    }
+
+    /**
+     *  @dev See {IIdentityRegistryStorage-addIdentityToStorage}.
+     */
+    function addIdentityToStorage() external view override onlyAgent {
```
Joachim-Lebrun, line 102: the whole idea of this contract is to provide a decentralized way of managing the links between OIDs and wallet addresses, therefore i think we should allow users to add and remove wallets linked to their OID, but to make sure they are really owning these wallets, they should provide a signature done by this wallet to make the transaction happen. the transaction should be called either by a wallet that is a management key on the OID or by the OID itself (through approve/execute)

contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
```diff
@@ -0,0 +1,173 @@
...
+     */
+    function modifyStoredInvestorCountry() external view override onlyAgent {
+        revert ContractIsReadOnly();
+
+    }
+
+    /**
+     *  @dev See {IIdentityRegistryStorage-removeIdentityFromStorage}.
+     */
+    function removeIdentityFromStorage() external view override onlyAgent {
```
Joachim-Lebrun, line 125: same comment as the addIdentity stuff, but in this case no need of a signature. A wallet can be removed by another wallet that is a management key on the OID or by the OID itself, note that you cannot remove the initial wallet linked to the OID on the IDFactory as that one is registered there forever

contracts/registry/implementation/IdentityRegistryStorageERC3643.sol
Joachim-Lebrun: call it GlobalIRS or something

PR #234, opened 2024-12-09
BT-385 Add InitialLockupPeriod module

pgonday:
https://tokeny.atlassian.net/browse/BT-385

### DIFF
--- a/contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
+++ b/contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
@@ -92,8 +92,13 @@ contract InitialLockupPeriodModule is AbstractModuleUpgradeable {
         uint256 releaseTimestamp;
     }
 
+    struct LockedDetails {
+        uint256 totalLocked;
+        LockedTokens[] lockedTokens;
+    }
+
     mapping(address compliance => uint256 lockupPeriod) private _lockupPeriods;
-    mapping(address compliance => mapping(address user => LockedTokens[])) private _lockedTokens;
+    mapping(address compliance => mapping(address user => LockedDetails)) private _lockedDetails;
 
     /// @dev initializes the contract and sets the initial state.
     function initialize() external initializer {
@@ -114,9 +119,9 @@ contract InitialLockupPeriodModule is AbstractModuleUpgradeable {
             return;
         }
 
-        (uint256 lockedAmount, uint256 unlockedAmount) = _calculateLockedAmount(msg.sender, _from);
+        LockedDetails storage lockedDetails =  _lockedDetails[msg.sender][_from];
         uint256 freeAmount = 
-            IToken(IModularCompliance(msg.sender).getTokenBound()).balanceOf(_from) - lockedAmount - unlockedAmount;
+            IToken(IModularCompliance(msg.sender).getTokenBound()).balanceOf(_from) - lockedDetails.totalLocked;
         if (_value > freeAmount) {
             _updateLockedTokens(_from, _value - freeAmount);
         }
@@ -124,7 +129,9 @@ contract InitialLockupPeriodModule is AbstractModuleUpgradeable {
 
     /// @inheritdoc IModule
     function moduleMintAction(address _to, uint256 _value) external override onlyComplianceCall {
-        _lockedTokens[msg.sender][_to].push(
+        LockedDetails storage lockedDetails =  _lockedDetails[msg.sender][_to];
+        lockedDetails.totalLocked += _value;
+        lockedDetails.lockedTokens.push(
             LockedTokens({
                 amount: _value,
                 releaseTimestamp: block.timestamp + _lockupPeriods[msg.sender]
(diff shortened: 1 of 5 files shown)


contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,161 @@
...
+    /// @dev initializes the contract and sets the initial state.
+    function initialize() external initializer {
+        __AbstractModule_init();
+    }
+
+    /// @dev sets the lockup period for a compliance contract.
+    /// @param _compliance the address of the compliance contract.
+    /// @param _lockupPeriod the lockup period in seconds.
+    function setLockupPeriod(address _compliance, uint256 _lockupPeriod) external onlyComplianceCall {
+        _lockupPeriods[_compliance] = _lockupPeriod;
```
SRL132, line 97: Just to make sure, is it fine that as long as the onlyComplianceCall check is satisfied, the sender can change the lockup period of any given compliance address? In modules such as `CountryRestrictModule.sol`, only restrictions for msg.sender can be added, per example.

pgonday: Oh yes I get your point, I will made an update, this should be:
```
function setLockupPeriod(uint256 _lockupPeriod) external onlyComplianceCall {
        _lockupPeriods[msg.sender] = _lockupPeriod;
```

SRL132: awesome, looks good to me

SRL132 (approved):

contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    mapping(address compliance => mapping(address user => LockedTokens[] lockedTokens)) private _lockedTokens;
+
+    /// @dev initializes the contract and sets the initial state.
+    function initialize() external initializer {
+        __AbstractModule_init();
+    }
+
+    /// @dev sets the lockup period for a compliance contract.
+    /// @param _lockupPeriod the lockup period in seconds.
+    function setLockupPeriod(uint256 _lockupPeriod) external onlyComplianceCall {
```
Joachim-Lebrun, line 95: probably more convenient for users to set a number of days, and the smart contract should transfor that in seconds as UNIX timestamp is calculated in seconds

contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,160 @@
...
+            LockedTokens({
+                amount: _value,
+                releaseTimestamp: block.timestamp + _lockupPeriods[msg.sender]
+            })
+        );
+    }
+
+    /// @inheritdoc IModule
+    // solhint-disable-next-line no-empty-blocks
+    function moduleBurnAction(address _from, uint256 _value) external override onlyComplianceCall { }
```
Joachim-Lebrun, line 124: need to manage the burning of locked tokens, otherwise you will have data on this module saying you have, for example 100 tokens unlocking in 3 days and 200 unlocking in 1 month while you only have 50 tokens left on your wallet. 
My opinion is that we need to burn first from the freely transferrable balance, and then from the balances that will be free the soonest. keeping the tokens that are going to unlock the latest untouched

contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    /// @inheritdoc IModule
+    function name() external pure override returns (string memory _name) {
+        return "InitialLockupPeriodModule";
+    }
+
+    /// @dev calculates the available amount of unlocked tokens for a user.
+    /// @param _compliance the address of the compliance contract.
+    /// @param _user the address of the user.
+    /// @return _availableAmount the available amount of unlocked tokens.
+    function _calculateAvailableAmount(address _compliance, address _user) internal view returns (uint256 _availableAmount) {
```
Joachim-Lebrun, line 151: rework this function with the totalLocked value as explained above. It should process the array of locked tokens only if there's not enough "already free" tokens to fulfill the needs of the current transaction

contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// @inheritdoc IModule
+    // solhint-disable-next-line no-empty-blocks
+    function moduleBurnAction(address _from, uint256 _value) external override onlyComplianceCall { }
+
+    /// @inheritdoc IModule
+    function moduleCheck(address _from, address /*_to*/, uint256 _value, address _compliance) external 
+        view override returns (bool) {
+        return _from == address(0) || _calculateAvailableAmount(_compliance, _from) >= _value;
```
Joachim-Lebrun, line 129: should call `_calculateAvailableAmount` only if balanceOf(_from) - _totalLocked < _value

contracts/compliance/modular/modules/InitialLockupPeriodModule.sol
```diff
@@ -0,0 +1,160 @@
...
+contract InitialLockupPeriodModule is AbstractModuleUpgradeable {
+
+    struct LockedTokens {
+        uint256 amount;
+        uint256 releaseTimestamp;
+    }
+
+    mapping(address compliance => uint256 lockupPeriod) private _lockupPeriods;
+    mapping(address compliance => mapping(address user => LockedTokens[] lockedTokens)) private _lockedTokens;
+
```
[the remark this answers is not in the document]
pgonday: I think this is not possible, as the locked tokens are block.timestamp dependant, they have to be calculated each time, or maybe with a more complex mechanism, which record the closest locked timestamp. But as the moduleCheck function is a view, it will not be possible to update the total at this level...
One possible optimisation is to delete locked tokens only if the free balance is not sufficient for transfer/burn.

Joachim-Lebrun (approved):

PR #236, opened 2024-12-16
BT-235: ERC2612 - Add ERC-20 Permit function compatibility

pgonday:
https://tokeny.atlassian.net/browse/BT-235


contracts/utils/EIP712Upgradeable.sol
```diff
@@ -0,0 +1,217 @@
...
+
+    bytes32 private constant _TYPE_HASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+
+    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.EIP712")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant _EIP712_STORAGE_LOCATION = 0xa16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d100;
+
+    /**
+     * @dev See {IERC-5267}.
+     */
```
SRL132, line 53: Is there a reason why this function is public and not external? It seems to be external in the specification  and I haven't seen it being explicitly called internally within the contract

https://eips.ethereum.org/EIPS/eip-5267

contracts/utils/EIP712Upgradeable.sol
```diff
@@ -0,0 +1,217 @@
...
+            uint256 chainId,
+            address verifyingContract,
+            bytes32 salt,
+            uint256[] memory extensions
+        )
+    {
+        EIP712Storage storage $ = _getEIP712Storage();
+        // If the hashed name and version in storage are non-zero, the contract hasn't been properly initialized
+        // and the EIP712 domain is not reliable, as it will be missing name and version.
+        require($._hashedName == 0 && $._hashedVersion == 0, "EIP712: Uninitialized");
```
SRL132, line 71: Custom error instead of string could help to save some gas here

contracts/utils/EIP712Upgradeable.sol
```diff
@@ -0,0 +1,217 @@
...
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+
+    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.EIP712")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant _EIP712_STORAGE_LOCATION = 0xa16a46d94261c7517cc8ff89f61c0ce93598e3c849801011dee649a6a557d100;
+
+    /**
+     * @dev See {IERC-5267}.
+     */
+    function eip712Domain()
+        public
```
SRL132, line 55: Is there a reason why this function is public and not external? It seems to be external in the specification and I haven't seen it being explicitly called internally within the contract

https://eips.ethereum.org/EIPS/eip-5267

contracts/token/TokenPermit.sol
```diff
@@ -0,0 +1,121 @@
...
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+pragma solidity 0.8.27;
+
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
```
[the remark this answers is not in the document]
pgonday: Yes, that as been done for token-factory, but in trex all import are "classical"... But ok to update it.

contracts/utils/EIP712Upgradeable.sol
pgonday: In fact I have just copy/paste OpenZeppelin contract to be sure to not add regression (just reorder function for solhint). At first I try to just import the file, but I haven't succeed to manage several version into the package.json. I need the last version of contract-upgradeable (for the custom storage), but it have dependencies on last OpenZeppelin contracts which is not the one we are using...
So either we find a way to just import the correct version, if not we can maybe replace with a custom error.

contracts/utils/EIP712Upgradeable.sol
pgonday: It's public in OZ implementation (https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v5.2/contracts/utils/cryptography/EIP712Upgradeable.sol) but yes it can be just external.

contracts/token/Token.sol
```diff
@@ -119,7 +120,13 @@ error DefaultAllowanceAlreadyDisabled(address _user);
 error DefaultAllowanceAlreadySet(address _target);
 
 
-contract Token is IToken, AgentRoleUpgradeable, TokenStorage, IERC165 {
+contract Token is IToken, AgentRoleUpgradeable, TokenStorage, IERC165, TokenPermit {
+    
+    bytes32 private constant _TYPE_HASH =
```
Joachim-Lebrun, line 125: we put all the variables and constants in the TokenStorage contract to separate storage from implementation and make sure we don't create collisions when we make an update. Otherwise if it is not practical to use the TokenStorage contract then you need to assign a specific storage slot to the data

SRL132 (approved):

contracts/token/TokenPermit.sol
```diff
@@ -0,0 +1,121 @@
...
+        return _domainSeparatorV4();
+    }
+
+     /// @inheritdoc IERC20Permit
+    function nonces(address owner) public view override(IERC20Permit, NoncesUpgradeable) returns (uint256) {
+        return super.nonces(owner);
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function __TokenPermit_init(string memory name) internal {
```
Joachim-Lebrun, line 114: need to find a solution without this init, because it has to work on updated contracts that didn't get initialized from the constructor. Or you can probably have something that checks if it is initialized already and calls __TokenPermit_init() in case it was not initialized at deployment

pgonday: In fact we can't do an init in Permit, because dapp/caller can use domainSeparator or eip712Domain to build the signature... So instead of duplicating name/version in the EIP712 part, it directly uses the name and version of the Token, and there is no need to further initialization.

SRL132 (approved):

Joachim-Lebrun (approved):

PR #244, opened 2025-03-03
Issue#242 perf: add allowance pre-check to transferFrom()

CT77777:
Add pre-check allowance before complicated identity and compliance verification.
It saves gas if transaction is reverted due to insufficient allowance.


Joachim-Lebrun:
@CT77777 did you compare the gas costs of the 2 implementations? because the purpose here is to save gas with the unchecked arithmetic operation, but given the implementation you propose i doubt it saves gas

